PR #3, opened 2022-04-15
Add eslint, remark, and retext linting

aduh95:
First time with MDX, it's… rough. Tools doesn't seem to nice with it, documentation is often sparse, it feels like fighting an uphill battle so far.


package.json
```diff
@@ -26,6 +30,13 @@
   "devDependencies": {
     "@docusaurus/module-type-aliases": "2.0.0-beta.17",
     "@tsconfig/docusaurus": "^1.0.4",
+    "eslint-config-prettier": "^8.5.0",
+    "eslint-plugin-mdx": "^1.17.0",
```
Murderlon, line 34: a new major version is out which perhaps improves things

package.json
```diff
@@ -2,6 +2,9 @@
   "name": "uppy-io",
   "version": "0.0.0",
   "private": true,
+  "lint-staged": {
+    "**/*": "prettier --write --ignore-unknown"
```
Murderlon, line 6: we want to let remark format our markdown and mdx instead of prettier

aduh95: Why not both?

Murderlon:
> We should, but I don't think it should hold this PR, IMO we better have some linting now and improve it bit by bit – don't let perfect be the enemy of good is what I'm trying to say.

I don't think it's in a state to merge. And most of our docs are MDX, so this linter setup won't help for all the contributions that are coming now, which are doc changes. It's also not like the improvements are days of work, but just a couple hours, so the impact of holding off is small

Murderlon:
Todo:

- Fix all warnings (80 or so)
- Setup GitHub Actions.

Could be done in separate PR?

Murderlon:
Update: fixed all warnings

PR #4, opened 2022-04-19
Fix Github actions preview deploy

arturi:
test


Murderlon:
Why do we have so many jobs? Couldn't it be one?

Capture d’écran 2022-04-19 à 12 04 49

PR #5, opened 2022-04-21
Rewrite Companion docs

Murderlon:
- [x] Big rewrite with endless tiny tweaks
- [x] Add guide how to scale companion (sticky sessions, trade-offs)
- [x] Explain the redis options in detail, what do they enable?

https://transloadit.github.io/uppy.io/pr-preview/pr-5/docs/companion


docs/companion.md
```diff
@@ -3,3 +3,693 @@ sidebar_position: 4
...
+#### `uploadUrls` `COMPANION_UPLOAD_URLS`
+
+An allowlist (array) of strings (exact URLs) or regular expressions.
+If specified, Companion will only accept uploads to these URLs.
+This is needed to make sure a Companion instance is only allowed to upload to your servers.
+
+:::caution
+Omitting this leaves your system open to potential [SSRF](https://en.wikipedia.org/wiki/Server-side_request_forgery) attacks,
+and may throw an error in future `@uppy/companion` releases.
+:::
```
mifi, line 256:
```suggestion
An allowlist (array) of strings (exact URLs) or regular expressions. Companion will only accept uploads to these URLs. This ensures that your Companion instance is only allowed to upload to your trusted servers and prevents [SSRF](https://en.wikipedia.org/wiki/Server-side_request_forgery) attacks. 
```

docs/companion.md
```diff
@@ -3,3 +3,693 @@ sidebar_position: 4
...
+#### `metrics` `COMPANION_HIDE_METRICS`
+ 
+A boolean flag to tell Companion whether to provide an endpoint `/metrics` with Prometheus metrics.
+
+#### `streamingUpload` `COMPANION_STREAMING_UPLOAD`
+
+A boolean flag to tell Companion whether to enable streaming uploads.
+If enabled, it will lead to _faster uploads_ because companion will start uploading at the same time as downloading using `stream.pipe`.
+If `false`, files will be fully downloaded first, then uploaded.
+Defaults to `false`.
```
mifi, line 438:
```suggestion
A boolean flag to tell Companion whether to enable streaming uploads. If enabled, it will lead to _faster uploads_ because companion will start uploading at the same time as downloading using `stream.pipe`. If `false`, files will be fully downloaded first, then uploaded. Defaults to `false`, but we recommended enabling it, especially if you're expecting to upload large files. In future versions the default might change to `true`.
```

docs/companion.md
```diff
@@ -3,3 +3,683 @@ sidebar_position: 4
...
+```
+
+Then run:
+
+```bash
+companion
+```
+
+You can also pass in the path to your JSON config file, like so:
+
```
mifi, line 152:
```suggestion

:::info
Companion will run as http and expects a reverse proxy with SSL termination in front of it when running in production. See `COMPANION_PROTOCOL` [option](#options) for more information.
:::
```

because someone was confused by this here: https://community.transloadit.com/t/companion-generates-redirect-urls-without-ssl/16264

docs/companion.md
```diff
@@ -3,3 +3,683 @@ sidebar_position: 4
...
+| ---------------------------------------- | ----------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `protocol` `COMPANION_PROTOCOL`          | `http` or `https` | Used to build a URL to reference the Companion instance itself, which is used for headers and cookies. Companion does not serve SSL certificates, so by default you should use `http`. You want to set this to `https` if you’ve enabled SSL/HTTPS for your domain by running a reverse https-proxy in front of Companion, or with a built-in HTTPS feature of your hosting service, such as Heroku.                                     |
```
mifi, line 280:
```suggestion
| `protocol` `COMPANION_PROTOCOL`          | `http` or `https` | Used to build a URL to reference the Companion instance itself, which is used for headers and cookies. Companion only supports running a HTTP server, however in production HTTPS is usually expected, so you would want to set this to `https` when running in production. You then need to set up a reverse proxy with SSL termination in front of Companion. Many cloud providers have built-in HTTPS termination.                                     |
```

because someone was confused by this here: https://community.transloadit.com/t/companion-generates-redirect-urls-without-ssl/16264

mifi (commented):
see https://community.transloadit.com/t/companion-generates-redirect-urls-without-ssl/16264

docs/companion.md
```diff
@@ -3,3 +3,683 @@ sidebar_position: 4
...
+
+companion.socket(server)
+```
+
+If WebSockets fail for some reason Uppy and Companion will fallback to HTTP polling.
+
+### Standalone
+
+You can use the standalone version if you want to run Companion as it’s own Node process.
+The standalone version is a configured Express server with sessions, logging, and security best practices.
```
Murderlon, line 128:
```suggestion
You can use the standalone version if you want to run Companion as it’s own Node process.
It is a configured Express server with sessions, logging, and security best practices.

Standalone Companion will always run as HTTP and expects a reverse proxy with SSL termination in front of it when running in production. See [`COMPANION_PROTOCOL`](#server) for more information.
```

docs/companion.md
Murderlon: I moved this up, see my other review comment

docs/companion.md
mifi: oh, but i'm not sure if we should recommend using `http`, because from a security standpoint, using http without SSL is strongly discouraged, when on the internet. Also not sure if we should recommend Heroku and give them free publicity

Murderlon: Good points! Updated the comment

mifi:
I have now added scaling docs from https://github.com/transloadit/uppy/pull/4227 and improved a bit. Please let me know if there's anything else I should improve :)

Murderlon:
Let's first hammer out the feedback in the other PR, and then copy it here again :)

docs/companion.md
```diff
@@ -3,3 +3,705 @@ sidebar_position: 4
...
+app.use(uppy.app({
+  providerOptions: {
+    s3: {
+      getKey: (req, filename, metadata) => filename,
+    },
+  },
+}))
+```
+
+#### `COMPANION_AWS_SECRET_FILE`
```
Murderlon, line 416: @mifi we need more info on how this is used and what the file should look like. Can you make a suggestion?

mifi: seems like it's not in use so i'll remove it

docs/companion.md
```diff
@@ -3,3 +3,716 @@ sidebar_position: 4
...
+#### `chunkSize` `COMPANION_CHUNK_SIZE`
+
+Controls how big the uploaded chunks are for AWS S3 Multipart and Tus.
+Smaller values lead to more overhead, but larger values lead to slower retries in case of bad network connections.
+Passed to tus-js-client [`chunkSize`](https://github.com/tus/tus-js-client/blob/master/docs/api.md#chunksize) as well as [AWS S3 Multipart](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/S3.html) `partSize`.
+
+### Standalone only options
+
+These environment variables apply only to Companion running in [standalone mode](#standalone).
+
```
mifi, line 496:
```suggestion

#### `REDIS_EXPRESS_SESSION_PREFIX`

Set a custom prefix for redis keys created by [connect-redis](https://github.com/tj/connect-redis). Defaults to `sess:`.  Sessions are used for storing authentication state and for allowing thumbnails to be loaded by the browser via Companion.

```

docs/companion.md
Murderlon: So this should be a relative path to a file with any extension which _only_ contains the secret directly, nothing else?

docs/companion.md
Murderlon: Could we add a small why you would want to change this?

docs/companion.md
```diff
@@ -3,3 +3,716 @@ sidebar_position: 4
...
+Settings this variable to `true` enables a route that can be used to validate your app with OneDrive. Only set this if you are sure that you are also setting the correct value for `COMPANION_ONEDRIVE_KEY` (i.e application ID - **not** the secret!)
+
+#### `COMPANION_HIDE_WELCOME`
+
+Setting this to `true` disables the welcome message shown at `/`.
+
+### Options `_FILE` suffix
+
+Many environment variables can alternatively be specified using a `_FILE` suffix. In this case the secret will be loaded from a file instead of being specified directly on the command line. This can be useful to prevent accidentally leaking secrets.
```
mifi, line 511:
```suggestion
Many environment variables can alternatively be specified using a `_FILE` suffix. In this case the secret will be loaded from a file instead of being specified directly on the command line. This can be useful to prevent accidentally leaking secrets. The should be an absolute path to a file with any extension which only contains the secret, nothing else.
```

docs/companion.md
```diff
@@ -490,28 +502,6 @@ Controls how big the uploaded chunks are for AWS S3 Multipart and Tus.
...
-
-#### `COMPANION_HIDE_WELCOME`
-
-Setting this to `true` disables the welcome message shown at `/`.
-
-### Options `_FILE` suffix
-
-Many environment variables can alternatively be specified using a `_FILE` suffix. In this case the secret will be loaded from a file instead of being specified directly on the command line. This can be useful to prevent accidentally leaking secrets.
-
-Here is a list of the suffixed options: `COMPANION_GOOGLE_SECRET_FILE`, `COMPANION_DROPBOX_SECRET_FILE`, `COMPANION_BOX_SECRET_FILE`, `COMPANION_INSTAGRAM_SECRET_FILE`, `COMPANION_FACEBOOK_SECRET_FILE`, `COMPANION_ONEDRIVE_SECRET_FILE`, `COMPANION_ZOOM_SECRET_FILE`, `COMPANION_ZOOM_VERIFICATION_TOKEN`, `COMPANION_AWS_SECRET_FILE`, `COMPANION_SECRET_FILE`, `COMPANION_PREAUTH_SECRET_FILE`
```
mifi: Shouldn't we document this somewhere?

docs/companion.md
```diff
@@ -3,3 +3,706 @@ sidebar_position: 4
...
+
+Like COMPANION_CLIENT_ORIGINS, but allows a single regex instead.
+`COMPANION_CLIENT_ORIGINS` will be ignored if this is used.
+
+#### `chunkSize` `COMPANION_CHUNK_SIZE`
+
+Controls how big the uploaded chunks are for AWS S3 Multipart and Tus.
+Smaller values lead to more overhead, but larger values lead to slower retries in case of bad network connections.
+Passed to tus-js-client [`chunkSize`](https://github.com/tus/tus-js-client/blob/master/docs/api.md#chunksize) as well as [AWS S3 Multipart](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/S3.html) `partSize`.
+
```
mifi, line 504:
```suggestion


Many environment variables can alternatively be specified using a `_FILE` suffix. In this case the secret will be loaded from a file instead of being specified directly on the command line. This can be useful to prevent accidentally leaking secrets.

Here is a list of the suffixed options: `COMPANION_GOOGLE_SECRET_FILE`, `COMPANION_DROPBOX_SECRET_FILE`, `COMPANION_BOX_SECRET_FILE`, `COMPANION_INSTAGRAM_SECRET_FILE`, `COMPANION_FACEBOOK_SECRET_FILE`, `COMPANION_ONEDRIVE_SECRET_FILE`, `COMPANION_ZOOM_SECRET_FILE`, `COMPANION_ZOOM_VERIFICATION_TOKEN`, `COMPANION_AWS_SECRET_FILE`, `COMPANION_SECRET_FILE`, `COMPANION_PREAUTH_SECRET_FILE`
```

docs/companion.md
Murderlon: They were already there, I just de-duplicated them

Screenshot 2022-12-27 at 12 40 43

mifi: I believe some of them are missing

Murderlon: Added some missing ones. I still have to document `COMPANION_AWS_SECRET_FILE` and `COMPANION_SECRET_FILE` but I believe they need more explanation because unlike the others, it seems to mean multiple vars per file? How does that need to look then?

mifi: Where do you see that they mean multiple vars per file?

Murderlon: somehow thought it would be a replacement for the S3 vars, such as the key. But it's only the secret there too of course. Added the last ones

mifi (approved):

PR #18, opened 2022-07-25
Fix documentation on retry-all vs upload-retry

ombr:
Hi,

This is related to https://github.com/transloadit/uppy/pull/3898

When a user do a retry-all no upload-retry event is triggered. (I had to debug this)
The retry-all event is not documented and I think behavior is unexpected (not triggering upload-retry at all)

This at least document current the behavior.

Thanks.

Luc

### DIFF
--- a/docs/uppy-core.mdx
+++ b/docs/uppy-core.mdx
@@ -1116,6 +1116,8 @@ uppy.on('upload-error', (file, error, response) => {
 #### `upload-retry`
 
 Fired when an upload has been retried (after an error, for example).
+  
+> ⚠️ Note that this method is event is not triggered when the user retries all uploads, it will trigger the `retry-all` event intstead.
 
 **Parameters**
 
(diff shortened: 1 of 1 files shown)


docs/uppy-core.mdx
```diff
@@ -1116,6 +1116,8 @@ uppy.on('upload-error', (file, error, response) => {
 #### `upload-retry`
 
 Fired when an upload has been retried (after an error, for example).
+  
+> ⚠️ Note that this method is event is not triggered when the user retries all uploads, it will trigger the `retry-all` event intstead.
```
Murderlon, line 1120:
```suggestion
:::note
Note that this method is event is not triggered when the user retries all uploads, it will trigger the `retry-all` event intstead.
:::
```

docs/uppy-core.mdx
```diff
@@ -1128,6 +1130,22 @@ uppy.on('upload-retry', (fileID) => {
   console.log('upload retried:', fileID)
 })
 ```
+  
+### `retry-all`
```
Murderlon, line 1134:
```suggestion
#### `retry-all`
```

Murderlon (commented):
Thanks for also making the PR here :)

PR #26, opened 2022-08-31
Add Transloadit docs

Murderlon:


docs/upload-strategies/transloadit.mdx
```diff
@@ -0,0 +1,585 @@
...
+#### `id`
+
+A unique identifier for this plugin (`String`, default: `'Transloadit'`).
+
+#### `service`
+
+The Transloadit API URL to use (`String`, default: `https://api2.transloadit.com`).
+
+The default will try to route traffic efficiently based on the location of your users.
+You could for instance set it to`https://api2-us-east-1.transloadit.com` if you want to use a particular region.
```
aduh95, line 163:
```suggestion
You could for instance set it to `https://api2-us-east-1.transloadit.com` if you want to use a particular region.
```

docs/upload-strategies/transloadit.mdx
```diff
@@ -0,0 +1,585 @@
...
+
+#### `limit`
+
+Limit the amount of uploads going on at the same time (`Number`, default: `5`).
```
aduh95, line 167: Same here
```suggestion
Limit the amount of uploads going on at the same time (`number`, default: `5`).
```

docs/upload-strategies/transloadit.mdx
```diff
@@ -0,0 +1,585 @@
...
+})
+```
+
+Form fields can also be computed dynamically using custom logic, by using the [`getAssemblyOptions(file)`](/docs/transloadit/#getAssemblyOptions-file) option.
+
+</details>
+
+### Static properties
+
+#### `Transloadit.COMPANION`
```
aduh95, line 441: This is deprecated, users should use `import {COMPANION_URL} from '@uppy/transloadit'`

aduh95 (changes requested):

aduh95:
Looking good. I feel very strongly that we should not use the capitalised version of types if we mean the non-capitalised one.

Murderlon:
> Looking good. I feel very strongly that we should not use the capitalised version of types if we mean the non-capitalised one.

Going to have to change all the docs then 🙈

aduh95:
Yes, what were you thinking ahah

Murderlon:
> Yes, what were you thinking ahah

I mean there were 15 previous PR's which used capitalized types but we're catching it now, which is okay, but it means it will be another PR for _all_ the docs.

docs/upload-strategies/transloadit.mdx
```diff
@@ -0,0 +1,585 @@
...
+  importFromUploadURLs: true,
+  assemblyOptions: {
+    params: {
+      auth: { key: 'YOUR_API_KEY' },
+      template_id: 'YOUR_TEMPLATE_ID',
+    }
+  },
+})
+```
+
```
kvz, line 334:
```suggestion
Tranloadit will download the files and expose them to your Template as `:original`, just as if they were directly uploaded from the Uppy client.  
```

docs/upload-strategies/transloadit.mdx
```diff
@@ -0,0 +1,585 @@
...
+#### `alwaysRunAssembly`
+
+Always create and run an Assembly when `uppy.upload()` is called, even if no files were selected (`Boolean`, default: `false`).
+
+This allows running Assemblies that do not receive files,
+but instead use a robot like [`/s3/import`](https://transloadit.com/docs/transcoding/#s3-import) to download
+the files from elsewhere, for example, for a bulk transcoding job.
+
+#### `locale`
+
```
kvz, line 350: This could use an introductory sentence maybe

docs/upload-strategies/transloadit.mdx
```diff
@@ -0,0 +1,585 @@
...
+Form fields can also be computed dynamically using custom logic, by using the [`getAssemblyOptions(file)`](/docs/transloadit/#getAssemblyOptions-file) option.
+
+</details>
+
+### Static properties
+
+#### `Transloadit.COMPANION`
+
+The main endpoint for Transloadit’s hosted companions. You can use this constant in remote provider options, like so:
```
kvz, line 443:
```suggestion
The main endpoint for Transloadit’s hosted Companions servers. You can use this constant in remote provider options, like so:
```

kvz (changes requested):
Left a few notes, you may not want to incorporate all of them. I think we should consistently capitalize Transloadit terminology. In most cases that's done, but I saw templates often being lowercased, didn't bother to create suggestions for each of them 👌 

Other than that, very nice work Merlijn! 💪

docs/upload-strategies/transloadit.mdx
Murderlon: This is a bigger todo, as this section will be generated.

docs/upload-strategies/transloadit.mdx
```diff
@@ -0,0 +1,585 @@
...
+  ```shell
+  yarn add @uppy/transloadit
+  ```
+
+  </TabItem>
+
+  <TabItem value="cdn" label="CDN">
+  <Admonition type="caution">
+    <p>
+      The bundle consists of most Uppy plugins, so this method is not recommended for production,
```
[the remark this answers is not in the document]
Murderlon: going to tackle this in a different PR as all docs need to change

PR #27, opened 2022-09-12
change to yarn 3

mifi:
also fix readme


.github/workflows/preview.yml
```diff
@@ -15,7 +15,12 @@ jobs:
     runs-on: ubuntu-20.04
     steps:
       - name: Checkout
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
+
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 16
```
aduh95, line 22:
```suggestion
          node-version: lts/*
```

.github/workflows/preview.yml
```diff
@@ -26,10 +31,10 @@ jobs:
 
       - name: Install and Build
         run: |
-          npm install
+          corepack yarn install --frozen-lockfile
```
aduh95, line 34: `--frozen-lockfile` is for Yarn 1.x, it's been deprecated on modern Yarn (Berry).
```suggestion
          corepack yarn install --immutable
```

.gitignore
```diff
@@ -20,3 +20,11 @@
 npm-debug.log*
 yarn-debug.log*
 yarn-error.log*
+
+.pnp.*
+.yarn/*
+!.yarn/patches
+!.yarn/plugins
+!.yarn/releases
```
aduh95, line 28: We're using Corepack, we don't need to commit Yarn. Can you also `git rm .yarn/releases/yarn-3.2.3.cjs` please?
```suggestion
```

.yarnrc.yml
```diff
@@ -0,0 +1,2 @@
+yarnPath: .yarn/releases/yarn-3.2.3.cjs
```
aduh95, line 1:
```suggestion
```

package.json
```diff
@@ -83,5 +84,6 @@
 		"typescript": "^4.6.2",
 		"unified": "^10.0.0",
 		"unified-message-control": "^4.0.0"
-	}
+	},
+	"packageManager": "yarn@3.2.3"
```
aduh95, line 88:
```suggestion
	"packageManager": "yarn@"3.2.3+sha224.953c8233f7a92884eee2de69a1b92d1f2ec1655e66d08071ba9a02fa"
```

aduh95 (changes requested):

.github/workflows/preview.yml
aduh95, line 22:
```suggestion
          node-version: lts/*
```

package.json
```diff
@@ -15,6 +15,7 @@
 		"remark:mdx": "remark docs -e mdx -u mdx -qf",
 		"serve": "docusaurus serve",
 		"start": "docusaurus start",
```
Murderlon, line 17:
```suggestion
```

mifi: this will break

package.json
mifi: what's the purpose of this hash?

aduh95:
```console
$ corepack yarn@3.2.1+sha224.deadbeef
Internal Error: Mismatch hashes. Expected deadbeef, got 953c8233f7a92884eee2de69a1b92d1f2ec1655e66d08071ba9a02fa
```

mifi: so it's an extra security layer to prevent accidentally downloading an attacker's yarn or a botched yarn release from the facebook team?

aduh95: > botched yarn release from the facebook team?

That would be extremely messed up as Facebook/Meta should have no control at all over that, they were only behind Yarn 1.x: https://yarnpkg.com/getting-started/qa#is-yarn-operated-by-facebook

> so it's an extra security layer to prevent accidentally downloading an attacker's yarn

Yes, it's a validation in case HTTPS is not enough / if the remote server was compromised.

aduh95 (approved):

Murderlon (approved):

Issue #15, opened 2022-07-25
Add docs versioning & automate Uppy version references


Murderlon:
Important read: https://docusaurus.io/docs/versioning
Note that the first version of this website when it's done would be the only historic version. We can't recreate from another structure, but that's fine, as of then we will have versioned docs.

Hooks like this could be used for references to versions in docs:

https://github.com/facebook/docusaurus/blob/13e7de853e0f0968916b60bae769bd49d224a19d/packages/docusaurus-plugin-content-docs/src/client/index.ts#L72

We need a smart way to have a single source of truth for things like this:

https://github.com/transloadit/uppy.io/blob/58eb8828bcb9d602374e9d166d707fff1cbdc8d7/docs/user-interfaces/dashboard.mdx?plain=1#L58

Inspiration how it's used:
https://github.com/facebook/docusaurus/blob/main/website/src/components/UpgradeGuide/index.tsx#L40=

Murderlon:
This may or may not depend on #14

mifi:
Not sure if we **really** want versioning. From their docs:

> :warning: Think about it before starting to version your documentation - it can become difficult for contributors to help improve it!

> Most of the time, you don't need versioning as it will just increase your build time, and introduce complexity to your codebase. Versioning is best suited for websites with high-traffic and rapid changes to documentation between versions. If your documentation rarely changes, don't add versioning to your documentation.

Do we really have rapid changes to docs between versions, that render docs incompatible? (other than majors)

Murderlon:
Fair point. I just thought it was nice majors but now that I think about it, it's much easier to use a migration guide and compare versions of docs. So we can skip this indeed

Murderlon:
Closed in #28

PR #28, opened 2022-09-12
reuse cdn sample code and version

mifi:
see #15

**don't merge before #27**


src/components/UppyCdnExample.tsx
```diff
@@ -0,0 +1,49 @@
...
+<link href="https://releases.transloadit.com/uppy/v${uppyVersion}/${uppyCssName}" rel="stylesheet">
+
+<!-- 2. Add JS before the closing \`</body>\` -->
+<script src="https://releases.transloadit.com/uppy/v${uppyVersion}/${uppyJsName}"></script>
+
+<!-- 3. Initialize -->
+<div id="uppy"></div>
+
+<script>
+  var uppy = new Uppy.Core()${linesProcessed.length > 0 ? `\n\n${linesProcessed}` : ''}
```
Murderlon, line 32: We shouldn't put uppy core here by default, I think it's nicer to see the complete example while writing docs and we may want to change the options for some examples.

src/components/UppyCdnExample.tsx
```diff
@@ -0,0 +1,49 @@
...
+export default function UppyCdnExample ({
+  children,
+  uppyCssName = 'uppy.min.css',
+  uppyJsName = 'uppy.min.js',
+}) {
+  let lines = [];
+
+  React.Children.toArray(children).forEach((child) => {
+    lines = [...lines, ...String(child).trim().split('\n').map((line) => line.trim())];
+  })
```
Murderlon, line 17: I'm surprised this is necessary, if we pass a correct string with newlines and everything, won't it look good?

Murderlon (commented):
Nice!

src/components/UppyCdnExample.tsx
```diff
@@ -0,0 +1,49 @@
+import React from 'react';
+import CodeBlock from '@theme/CodeBlock';
+import Admonition from '@theme/Admonition';
+
+
+const uppyVersion = '2.9.0' // todo
```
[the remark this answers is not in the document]
Murderlon: Note that we may already have these hooks in our deps

src/components/UppyCdnExample.tsx
mifi: What should we use instead? I see all the existing examples use Uppy.Core.

mifi: Oh you mean we should leave the script tag empty

src/components/UppyCdnExample.tsx
mifi: Yes, but I think it's nice if the developer doesn't have to worry about providing the correct indentation everytime we want to use the component.

src/components/UppyCdnExample.tsx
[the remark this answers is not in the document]
mifi: Couldn't get that hook to work but I used `import packageJson from '../../package.json'` and it seems to work

src/components/UppyCdnExample.tsx
Murderlon: No I meant it feels less magical to explicitly write an example every time, so I personally wouldn't put the default uppy core in here already

docs/sources/companion-plugins/url.mdx
```diff
@@ -42,27 +43,14 @@ so you don’t have to worry about the security implications of arbitrary URLs.
...
-      const Url = Uppy.Url
-    </script>
-  ```
-
+    <UppyCdnExample>
+      {`
+        var uppy = new Uppy.Core()
+        uppy.use(Uppy.Url, {
+          // Options
+        })
```
Murderlon, line 51: Everything seems to have an extra newline in between. Is the output different?

Murderlon (approved):

docs/sources/companion-plugins/url.mdx
Murderlon: Never mind it was GitHub's weird diff

aduh95:
`Uppy.Core` no longer exists in Uppy 3.x, now it's `Uppy.Uppy`.

Murderlon:
I forgot indeed that we now write out CDN example like so (`type="module"` and import from URL):

```js
<link href="https://releases.transloadit.com/uppy/v3.0.1/uppy.min.css" rel="stylesheet">

<div id="uppy"></div>

<script type="module">
  import { Uppy, Dashboard, Tus } from "https://releases.transloadit.com/uppy/v3.0.1/uppy.min.mjs"

  const uppy = new Uppy()
  uppy.use(Dashboard, { target: '#files-drag-drop' })
  uppy.use(Tus, { endpoint: 'https://tusd.tusdemo.net/files/' })
</script>
```

@mifi not sure how we can easily reference the correct version when we still write it everywhere instead of in the component?

mifi:
Not sure, but maybe we need to use some placeholder for interpolation? something like `v{{UPPY_VERSION}}` and do a `String(children).replace(/{{UPPY_VERSION}}/g, uppyVersion)`

Murderlon:
Feels a bit dirty but that might indeed be the only alternative to running a script which does the same.

Murderlon:
Would you be willing to give that a try? :)

PR #29, opened 2022-09-15
Add Tus docs

Murderlon:
https://transloadit.github.io/uppy.io/pr-preview/pr-29/docs/upload-strategies/tus


docs/upload-strategies/tus.mdx
```diff
@@ -0,0 +1,273 @@
...
+
+## When should I use it?
+
+:::tip
+Not sure which upload strategy is best for you? Read “[Choosing the upload strategy you need](/docs/guides/choosing-upload-strategy)”.
+:::
+
+[Tus][tus] is an open protocol for resumable uploads built on HTTP.
+This means accidentally closing your tab or losing connection let’s you continue, for instance, your 10GB upload
+instead of starting all over.
```
mifi, line 21:
```suggestion
This means that accidentally closing your tab or losing connection lets you continue, for instance, your 10GB upload instead of starting all over.
```

docs/upload-strategies/tus.mdx
```diff
@@ -0,0 +1,273 @@
...
+
+It supports any language, any platform, and any network. You can checkout the [implementations](https://tus.io/implementations.html) to
+find the implementation in your favorite language. You can store files on your self-hosted Tus server, but also use service integrations (such as S3) to store files
+externally.
```
mifi, line 25:
```suggestion
Tus consists of a client and a server. Multiple client and server [implementations](https://tus.io/implementations.html) exist in many languages. Uppy uses `tus-js-client` internally and you can set up any server you like to use with Uppy. You can store files on your self-hosted Tus server itself, but more common is to use service integrations (such as S3) to store files externally.

:::tip
If you don't want to host your own Tus server, you can take advantage of Transloadit's hosted services - read “[Choosing the upload strategy you need](/docs/guides/choosing-upload-strategy)”.
:::
```

Murderlon: I'm not sure we need to reiterate this here. At the end of the day, this is still an OSS product and the website as well. The guide is already linked 10 lines up and the fact that TL has hosted servers is also mentioned in FAQ. What do you think??

Murderlon: I would be more open to this change if we don't use `:::tip` and we don't link to the same thing as the previous tip does. Perhaps this can be one extra line, with a link to the FAQ below?

mifi: good idea! i updated the comment

Murderlon: I think I still prefer the previous text and would only add this new sentence:

```suggestion
It supports any language, any platform, and any network. You can checkout the [implementations](https://tus.io/implementations.html) to
find the implementation in your favorite language. You can store files on your self-hosted Tus server, but also use service integrations (such as S3) to store files
externally. If you don't want to host your own Tus server, you can take advantage of [Transloadit's hosted services](#are-there-hosted-tus-servers). 
```

mifi: I thought it would be worth mentioning that Tus is a server and a client, because newcomers might be confused as to why they would choose tus in any language when clearly Uppy is a Javascript based library. Whether Tus client supports many different languages/platforms is not so interesting for someone who just wants to get Uppy up and running.

Murderlon: > newcomers might be confused as to why they would choose tus in any language when clearly Uppy is a Javascript based library

People first need to setup a Tus server before they can do anything. Just because Uppy is JS, doesn't mean that their backend is, they may have an existing .NET backend. I think it's important to reference the implementations to choose from.

mifi: yes, that's why i want to mention that tus is a client and a server, so that they understand that they must choose a tus server, not a client

Acconut: I agree with @mifi here. The docs should mention that you need a tus server to use the protocol and plugin for uploading. For of the time, this is clear. But from time to time users still get confused.

docs/upload-strategies/tus.mdx
```diff
@@ -0,0 +1,273 @@
...
+  .use(Dashboard, { inline: true, target: 'body' })
+  .use(Tus, { endpoint: 'https://tusd.tusdemo.net/files/' })
+```
+
+## API
+
+### Options
+
+:::info
+All options are passed to `tus-js-client` and we document the ones here that we added or changed.
```
Acconut, line 80: If you say that you only mention options differing from tus-js-client, then why does it mention endpoint,, overridePatchMethod, and retryDelays?

docs/upload-strategies/tus.mdx
Murderlon: I incorporated your suggestions but made it a bit more succinct. This is the "Why should you use it section" and some details (like tus-js-client) are already provided in the section above.

```suggestion
Tus supports any language, any platform, and any network. It requires a client and server integration to work. You can checkout the client and server [implementations](https://tus.io/implementations.html) to
find the server in your preferred language. You can store files on the Tus server itself, but you can also use service integrations (such as S3) to store files externally. If you don't want to host your own server, see “[Are there hosted Tus servers?](#are-there-hosted-tus-servers)”.
```

docs/upload-strategies/tus.mdx
Murderlon: Good point! I should reword that we also document required options, like endpoint. The `retryDelays` default is different than the tus-js-client default. `overridePatchMethod` can indeed be removed.

Murderlon:
@Acconut do you have any additional FAQ entries we could add? Perhaps something about the cost of changing or moving fiels in hooks?

docs/upload-strategies/tus.mdx
```diff
@@ -0,0 +1,271 @@
...
+
+#### `withCredentials`
+
+Configure the requests to send Cookies using the [`xhr.withCredentials`](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/withCredentials) property (`boolean`, default: `false`).
+
+The remote server must accept CORS and credentials.
+
+#### `retryDelays`
+
+When uploading a chunk fails, automatically try again after the defined millisecond intervals (`array<number>`, default: `[0, 1000, 3000, 5000]`).
```
aduh95, line 139:
```suggestion
When uploading a chunk fails, automatically try again after the defined millisecond intervals (`Array<number>`, default: `[0, 1000, 3000, 5000]`).
```
or
```suggestion
When uploading a chunk fails, automatically try again after the defined millisecond intervals (`number[]`, default: `[0, 1000, 3000, 5000]`).
```

docs/upload-strategies/tus.mdx
```diff
@@ -0,0 +1,271 @@
...
+    if (res.getStatus() === 401) {
+      await refreshAuthToken()
+    }
+  },
+})
+```
+
+#### `allowedMetaFields`
+
+Pass an array of field names to limit the metadata fields that will be added to uploads as [Tus Metadata](https://tus.io/protocols/resumable-upload.html#upload-metadata) (`array`, default: `null`).
```
aduh95, line 183:
```suggestion
Pass an array of field names to limit the metadata fields that will be added to uploads as [Tus Metadata](https://tus.io/protocols/resumable-upload.html#upload-metadata) (`Array`, default: `null`).
```

docs/upload-strategies/tus.mdx
```diff
@@ -0,0 +1,271 @@
...
+
+If you want to change the file names, you want to do that in [`onBeforeFileAdded`](/docs/uppy-core#onbeforefileaddedfile-files).
+
+If you want to send extra headers with the request, use [`headers`](#headers) or [`onBeforeRequest`](#onbeforerequestreq-file).
+
+### How do I change (or move) files after sending them?
+
+If you want to preserve files names, extract meta data, or move files to a different place you generally can with hooks or events.
+It depends on the Tus server you use how it’s done exactly. [tusd][], for instance, exposes [hooks](https://github.com/tus/tusd/blob/master/docs/hooks.md)
+and [tus-node-server](https://github.com/tus/tus-node-server) has [events](https://github.com/tus/tus-node-server#events).
```
aduh95, line 238:
```suggestion
and [`tus-node-server`](https://github.com/tus/tus-node-server) has [events](https://github.com/tus/tus-node-server#events).
```

docs/upload-strategies/tus.mdx
```diff
@@ -0,0 +1,271 @@
...
+### How do I change files before sending them?
+
+If you want to change the file names, you want to do that in [`onBeforeFileAdded`](/docs/uppy-core#onbeforefileaddedfile-files).
+
+If you want to send extra headers with the request, use [`headers`](#headers) or [`onBeforeRequest`](#onbeforerequestreq-file).
+
+### How do I change (or move) files after sending them?
+
+If you want to preserve files names, extract meta data, or move files to a different place you generally can with hooks or events.
+It depends on the Tus server you use how it’s done exactly. [tusd][], for instance, exposes [hooks](https://github.com/tus/tusd/blob/master/docs/hooks.md)
```
aduh95, line 237:
```suggestion
It depends on the Tus server you use how it’s done exactly. [`tusd`][], for instance, exposes [hooks](https://github.com/tus/tusd/blob/master/docs/hooks.md)
```

docs/upload-strategies/tus.mdx
```diff
@@ -0,0 +1,271 @@
...
+and [tus-node-server](https://github.com/tus/tus-node-server) has [events](https://github.com/tus/tus-node-server#events).
+
+### Which server do you recommend?
+
+[Transloadit](https://transloadit.com) runs [tusd][] in production, where it serves millions of requests globally.
+Therefor we recommend tusd as battle-tested from our side, but other companies have had success with other [implementations][] so it depends on your needs.
+
+### Are there hosted Tus servers?
+
+All [Transloadit plans](https://transloadit.com/pricing) come with a hosted [tusd][] server.
```
aduh95, line 247:
```suggestion
All [Transloadit plans](https://transloadit.com/pricing) come with a hosted [`tusd`][] server.
```

docs/upload-strategies/tus.mdx
```diff
@@ -0,0 +1,271 @@
...
+
+First and foremost: reliable, resumable uploads. 
+This means accidentally closing your tab or losing connection let’s you continue, for instance, your 10GB upload
+instead of starting all over.
+
+Tus is also efficient with lots of files (such as 8K) and large files.
+Uploading to AWS S3 directly from the client also introduces quite a bit of overhead,
+as more requests are needed for the flow to work.
+
+[tus-js-client]: https://github.com/tus/tus-js-client
```
aduh95, line 263:
```suggestion
[`tus-js-client`]: https://github.com/tus/tus-js-client
```

docs/upload-strategies/tus.mdx
```diff
@@ -0,0 +1,271 @@
...
+
+[Uppy file]: /docs/uppy-core#working-with-uppy-files
+
+[tus]: https://tus.io/
+
+[tusd]: https://github.com/tus/tusd
```
aduh95, line 269:
```suggestion
[`tusd`]: https://github.com/tus/tusd
```

Acconut:
> Perhaps something about the cost of changing or moving fiels in hooks?

Sorry, you read that now. What exactly do you mean by this? I don't understand this question.

Murderlon:
We had a customer who was worried about the extra costs of having to change files after they have been send. Perhaps a QA on why that process is cheap in S3? Or other QA's you can think of?

PR #31, opened 2022-09-15
Fix casing of types in docs

aduh95:
Typically, the capitalised version (e.g. `String`) refers to instances of the class (e.g. `new String`), when in fact the docs mean to refer to the primitive type (e.g. `'a primitive string'`).

### DIFF
--- a/docs/sources/companion-plugins/box.mdx
+++ b/docs/sources/companion-plugins/box.mdx
@@ -125,19 +125,19 @@ companion.app({
 
 #### `id`
 
-A unique identifier for this plugin (`String`, default: `'Box'`).
+A unique identifier for this plugin (`string`, default: `'Box'`).
 
 #### `title`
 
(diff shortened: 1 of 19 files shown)


Murderlon (commented):
Thanks. Do we want to do the same for `Array`, `Object`, `Function`, `Element`, and `Regex`?

aduh95:
> Do we want to do the same for `Array`, `Object`, `Function`, `Element`, and `Regex`?

No, they are not primitive types, see https://developer.mozilla.org/en-US/docs/Glossary/Primitive.

`Object` is kinda misleading as we don't care if it's an instance of `Object` or not, but also there's no good alternatives and we export `.d.ts` that should contains a more useful types – same thing for `Function`.
`Regex` is a typo for `RegExp`, worth fixing indeed.

Murderlon (approved):

PR #32, opened 2022-09-15
try to explain upload strategies better

mifi:
I'm not sure why the change from "destination" to "upload strategy" was made, but I feel like we should explain that it's like a destination somehow, because it pairs well with "sources" in order to understand how it works for newbies

### DIFF
--- a/docs/guides/choosing-upload-strategy.md
+++ b/docs/guides/choosing-upload-strategy.md
@@ -4,7 +4,7 @@ sidebar_position: 1
 
 # Choosing the upload strategy you need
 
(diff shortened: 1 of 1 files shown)


mifi:
Relative to Uppy, I would say they are all destinations. Maybe not always a final destination, but you can say the same about S3 or XHR. It might be that a process will pick up the S3 or XHR uploads and process them further after the upload is done. Tus can also be a final destination, or it can be an intermediary. But seen from Uppy's point of view, it's where the file is destined to, IMO :)

Murderlon:
Closing this as we will be renaming "Upload strategies"

PR #35, opened 2022-09-20
Add AWS S3 docs

Murderlon:
https://transloadit.github.io/uppy.io/pr-preview/pr-35/docs/upload-strategies/aws-s3

### DIFF
--- a/docs/guides/choosing-upload-strategy.md
+++ b/docs/guides/choosing-upload-strategy.md
@@ -56,13 +56,13 @@ Uppy has two plugins to make this happen [`@uppy/aws-s3`][] and [`@uppy/aws-s3-m
 
 If your users are planning to mostly upload small files and/or a lot of files, it’s better to use [`@uppy/aws-s3`][].
 
-[`@uppy/aws-s3-multipart`][] starts to become valuable for bigger files (100MB+) as it uploads a single object as a set of parts.
-This has some benefits, such as improved throughput (uploading parts in parallel) and quick recovery from network issues (only the failed parts need to be retried).
-The downside is request overhead, as it needs to do setup, signing, and completion besides the upload requests.
-For example, if you are uploading files that are only a couple KB with a 100ms roundtrip latency,
-you are spending 400ms on overhead and only a couple milliseconds on uploading. 
+[`@uppy/aws-s3-multipart`][] starts to become valuable for larger files (100 MB+) as it uploads a single object as a set of parts.
+This has certain benefits, such as improved throughput (uploading parts in parallel) and quick recovery from network issues (only the failed parts need to be retried).
+The downside is request overhead, as it needs to do creation, signing, and completion requests besides the upload requests.
+For example, if you are uploading files that are only a couple kilobytes with a 100ms roundtrip latency,
+you are spending 400ms on overhead and only a few milliseconds on uploading. 
 
-If you are uploading big files (100MB+), we recommend [`@uppy/aws-s3-multipart`][], otherwise [`@uppy/aws-s3`][].
+If you are uploading large files (100 MB+), we recommend [`@uppy/aws-s3-multipart`][], otherwise [`@uppy/aws-s3`][].
 
 :::info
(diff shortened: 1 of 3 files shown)


docs/upload-strategies/aws-s3.mdx
```diff
@@ -0,0 +1,403 @@
...
+
+```js
+uppy.on('upload-success', (file, data) => {
+  const s3Key = file.meta['key'] // the S3 object key of the uploaded file
+})
+```
+
+### What makes a provider S3 compatible?
+
+...
```
Murderlon, line 401: @mifi do you know a good answer to this perhaps?

mifi: I think this has been answered in the existing docs:

https://uppy.io/docs/aws-s3/#S3-alternatives

although I think we should add a note saying something like we don’t officially support and/or test against any of these s3-providers, so it’s not a first-class support, but it should work. (I think I remember issues with s3-like providers before, where we decided not to implement special code to make it work with those)

Murderlon: That doesn't answer this question though, I mean how do you know if random provider X is S3 compatible? What does it take to be S3 compatible?

Regarding the note, that's already in "when should I use this?" section

docs/upload-strategies/aws-s3.mdx
```diff
@@ -0,0 +1,403 @@
...
+*   Set this to `null` (the default) to send _all_ metadata fields.
+*   Set this to an empty array `[]` to not send any fields.
+
+#### `getUploadParameters(file)`
+
+:::note
+When using [Companion][companion docs] to sign S3 uploads, do not define this option.
+:::
+
+A function that returns upload parameters for a file (`Promise<Object>`, default: `null`).
```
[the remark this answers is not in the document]
Murderlon: This is too detailed in my opinion, the structure is defined below and this is just a high level overview for everyone (so also people who don't know TS).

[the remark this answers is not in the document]
aduh95: Makes sense, in this case, let's remove `<Object>`.
```suggestion
A function that returns upload parameters for a file (`Promise`, default: `null`).
```

docs/upload-strategies/aws-s3.mdx
mifi: I don't really have a good answer to this question. It would require a lot of testing to see if all functionality that Uppy needs is supported by each s3-like provider, also at scale. Or we can just say that we don't provide a list of supported s3-like providers, but we treat s3 with duck-typing in mind. "If it talks like s3 and it uploads like s3, then it's probably s3."

docs/guides/choosing-upload-strategy.md
```diff
@@ -56,38 +53,38 @@ Uppy has two plugins to make this happen [`@uppy/aws-s3`][] and [`@uppy/aws-s3-m
 
 If your users are planning to mostly upload small files and/or a lot of files, it’s better to use [`@uppy/aws-s3`][].
 
-[`@uppy/aws-s3-multipart`][] starts to become valuable for bigger files (100MB+) as it uploads a single object as a set of parts.
-This has some benefits, such as improved throughput (uploading parts in parallel) and quick recovery from network issues (only the failed parts need to be retried).
-The downside is request overhead, as it needs to do setup, signing, and completion besides the upload requests.
-For example, if you are uploading files that are only a couple KB with a 100ms roundtrip latency,
-you are spending 400ms on overhead and only a couple milliseconds on uploading. 
+[`@uppy/aws-s3-multipart`][] starts to become valuable for larger files (100&nbsp;MB+) as it uploads a single object as a set of parts.
```
mifi, line 56:
```suggestion
[`@uppy/aws-s3-multipart`][] is valuable for larger files (100&nbsp;MB+) as it uploads a single object as a set of parts.
```

mifi (approved):

docs/upload-strategies/aws-s3.mdx
Murderlon:
```suggestion
```

PR #43, opened 2022-12-22
@uppy/xhr-upload: add `'upload-stalled'` event

aduh95:
Refs: https://github.com/transloadit/uppy/pull/4247


Murderlon (approved):
Thanks for making the change here too

arturi:
Not sure we should have merged here before merging the PR in the main repo — what if we reject it.

PR #48, opened 2022-12-28
fix Prettier integration

aduh95:
- created a symlink between `.gitignore` and `.prettierignore` so both are synced.
- extended the doc linting to files outside of `docs/` so e.g. the blog and the README are also linted.
- added prettier to the lint-stage config as previously it was not due to how `lint-stage` handles files that fit several rules.
- clean up the list of npm scripts and use alphabetical order.


blog/2019-05-29-long-blog-post.md
```diff
@@ -5,40 +5,40 @@ authors: endi
 tags: [hello, docusaurus]
 ---
 
-This is the summary of a very long blog post,
+This is the summary of a v\*ry long blog post,
```
Murderlon, line 8: Where are all these escapes coming from?

Murderlon (commented):
Nice!

blog/2019-05-29-long-blog-post.md
aduh95: `very` is a word rejected by remark-simplify, so I'm using an asterisk to hide that naughty word. The backslash is added by Prettier so it's clear that asterisk doesn't delimit a `<em>` section. Not that it really matters since I expect we plan to use Uppy blog files instead of the current ones that are there just as a filler.

Murderlon (approved):

PR #49, opened 2022-12-28
add linter in GHA to check PRs and `main`

aduh95:
This should help catch instances where Prettier wasn't run on the commit hook for some reason.

### DIFF
--- a/.github/workflows/linters.yml
+++ b/.github/workflows/linters.yml
@@ -1,0 +1,47 @@
+name: Linters
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
+  push:
+    branches:
+      - main
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
+env:
+  NODE_VERSION: lts/*
+
+permissions:
+  contents: read
+
+jobs:
+  lint-js-and-md:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          persist-credentials: false
+      - name: Use Node.js ${{ env.NODE_VERSION }}
+        uses: actions/setup-node@v3
+        with:
+          node-version: ${{ env.NODE_VERSION }}
+      - name: Environment Information
+        run: npx envinfo
+      - name: Get yarn cache directory path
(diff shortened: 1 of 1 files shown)


.github/workflows/linters.yml
```diff
@@ -0,0 +1,47 @@
...
+      - uses: actions/cache@v3
+        id: yarn-cache # use this to check for `cache-hit` (`steps.yarn-cache.outputs.cache-hit != 'true'`)
+        with:
+          path: ${{ steps.yarn-cache-dir-path.outputs.dir }}
+          key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-yarn-
+      - name: Install dependencies
+        run: corepack yarn
+      - name: Lint JavaScript files
```
Murderlon, line 46:
```suggestion
      - name: Lint JavaScript and Markdown files
```

Murderlon (approved):

.github/workflows/linters.yml
```diff
@@ -0,0 +1,47 @@
...
+        uses: actions/setup-node@v3
+        with:
+          node-version: ${{ env.NODE_VERSION }}
+      - name: Environment Information
+        run: npx envinfo
+      - name: Get yarn cache directory path
+        id: yarn-cache-dir-path
+        run: echo "dir=$(corepack yarn config get cacheFolder)" >> $GITHUB_OUTPUT
+
+      - uses: actions/cache@v3
```
mifi, line 37: I dont think this is needed as caching is built into setup action
https://github.com/actions/setup-node#caching-global-packages-data

.github/workflows/linters.yml
aduh95: It also lints YAML, let’s simplify this. 
```suggestion
      - name: Lint files
```

.github/workflows/linters.yml
aduh95: I copied it from Uppy repo. IIRC we had to add this step because the cache only worked for Yarn v1, are we sure it’s going to work on more modern versions? I’d rather keep the step even if it turns out to be redundant than wasting time on every run because caching doesn’t work.

mifi (approved):

PR #58, opened 2023-01-17
Port `@uppy/remote-sources` docs

aduh95:


Murderlon (commented):
Nice start 👍 

Some changes required:
- We have never talked on the website about "preset plugins", so no one knows what that means. Could use a sentence what presets in uppy mean.
- Missing the "When should I use this?" section
- Code sample should not be under first heading but under "Use". You also forget the `target` option in the sample I think? `showLineNumbers` also required. CSS imports are missing.
- No tip admonition to a codesandbox
- "Installation" -> "Install"

aduh95:
> * We have never talked on the website about "preset plugins", so no one knows what that means. Could use a sentence what presets in uppy mean.

Note that the page in inside the "Presets" section in the nav menu: 
image
 
I'm not sure what there is to explain, the name seems quite self-explanatory. What would you suggest as an explaination?

> * You also forget the `target` option in the sample I think?

It defaults to `Dashboard`, so it wouldn't be necessary for this example to work. But I suppose we can always pass it explicitly. wdyt?

> * No tip admonition to a codesandbox

Not sure I follow, what do you mean?

docs/presets/remote-sources.mdx
```diff
@@ -1 +1,143 @@
...
+---
+
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+import UppyCdnExample from '/src/components/UppyCdnExample';
+
 # Remote sources
+
+`@uppy/remote-sources` is a preset plugin to add all the available remote
```
arturi, line 12:
```suggestion
`@uppy/remote-sources` is a preset plugin (meaning it bundles and sets up other plugins) to add all the available remote
```

arturi: Although that is explained right after.

docs/presets/remote-sources.mdx
```diff
@@ -1 +1,143 @@
...
+
+Remote Sources requires Dashboard and automatically installs all its plugins to
+it.
+
+:::
+
+## When should I use it?
+
+`@uppy/remote-sources` aims to simplify the setup for adding Companion plugins,
+when you want to share the configuration between plugins. If you want your user
```
arturi, line 26:
```suggestion
when you want to share the configuration between plugins. If you want your users
```

docs/presets/remote-sources.mdx
```diff
@@ -1 +1,143 @@
...
+to upload files from any remote source, this plugin is recommended.
```
arturi, line 27:
```suggestion
to upload files from any of the remote sources that Uppy offers, this plugin is recommended.
```

docs/presets/remote-sources.mdx
```diff
@@ -1 +1,143 @@
...
+
+A [Companion](/docs/companion) instance is required for the Facebook plugin to
```
arturi, line 29:
```suggestion
A [Companion](/docs/companion) instance is required for the Remote Sources plugin to
```

docs/presets/remote-sources.mdx
```diff
@@ -1 +1,143 @@
...
+work. Companion handles authentication with Facebook, downloads the files, and
```
arturi, line 30:
```suggestion
work. Companion handles authentication with Remote Sources, such as Facebook, Dropbox, etc, downloads the files, and
```

arturi: Add link to Companion docs? Probably also to the Dashboard above.

docs/presets/remote-sources.mdx
```diff
@@ -1 +1,143 @@
...
+uploads them to the destination. This saves the user bandwidth, especially
+helpful if they are on a mobile connection.
+
+You can self-host Companion or get a hosted version with any
+[Transloadit plan](https://transloadit.com/pricing/).
+
+## Install
+
+This plugin is published as the `@uppy/remote-sources` package on the npm
```
arturi, line 39:
```suggestion
```

arturi: We don't use that intro anywhere else in this section, I think.

Issue #45, opened 2022-12-23
Generate table of contents under "Options" heading for all docs


Murderlon:
The sidebar with links for every docs page is convenient to quickly navigate:

<img width="226" alt="Screenshot 2022-12-23 at 12 46 38" src="https://user-images.githubusercontent.com/9060226/209331118-87e68680-057f-482d-89b2-1a236de03fc0.png">

However it doesn't contain all options. Instead it would be nice to [generate](https://github.com/remarkjs/remark-toc/) a list of links of all the options under the "Options" heading in the page (so not in the sidebar).

kvz:
Would be nice to have post-launch

Murderlon:
This is blocked by https://github.com/facebook/docusaurus/issues/5379

PR #59, opened 2023-01-17
Port `@uppy/drop-target` docs

aduh95:


Murderlon (commented):
Nice, thanks.

Some changes still needed:

- It's missing the "When should I use it? section
- Code sample is not under first heading but under "Use"
- "Use" doesn't have a "CSS" section, instead if shows the plugin in context of other plugins with highlighting and `showLineNumbers`. Checkout some of the "Sources" plugins for examples.
- "Options" doesn't show a code overview of the options. We have #45 for that.
- Options do not show default values in the heading
- The tip admonition for examples doesn't exist. The codesandbox link is also incorrect

docs/user-interfaces/elements/drop-target.mdx
```diff
@@ -1 +1,104 @@
...
+The `@uppy/drop-target` plugin lets your users drag-and-drop files on any
+element on the page, for example the whole page, `document.body`.
+
+Can be used together with Uppy Dashboard or Drag & Drop plugins, or your custom
+solution, including plain text “please drop files here”.
+
+:::tip
+
+[Try out the live example](/examples) or take it for a spin in
+[CodeSandbox](https://codesandbox.io/s/uppy-drag-drop-gyewzx?file=/src/index.js).
```
Murderlon, line 21: /examples doesn't exit. We should also create a codesandbox for this plugin and link it here.

docs/user-interfaces/elements/drop-target.mdx
```diff
@@ -1 +1,104 @@
...
+
+import UppyCdnExample from '/src/components/UppyCdnExample';
+
 # Drop target
+
+The `@uppy/drop-target` plugin lets your users drag-and-drop files on any
+element on the page, for example the whole page, `document.body`.
+
+Can be used together with Uppy Dashboard or Drag & Drop plugins, or your custom
+solution, including plain text “please drop files here”.
```
Murderlon, line 16:
```suggestion
solution targeting any DOM element.
```

docs/user-interfaces/elements/drop-target.mdx
```diff
@@ -1 +1,104 @@
...
+:::
+
+## When should I use this?
+
+When you want to allow users to drag and drop files in your own UI, rather than
+in the [`Dashboard`](../dashboard) UI.
+
+## Install
+
+This plugin is published as the `@uppy/drop-target` package on the npm registry.
```
Murderlon, line 32:
```suggestion
```

docs/user-interfaces/elements/drop-target.mdx
```diff
@@ -1 +1,104 @@
...
+Event listener for the [`dragover` event][].
+
+#### `onDrop(event)`
+
+Event listener for the [`drop` event][].
+
+#### `target`
+
+DOM element, CSS selector, or plugin to place the drag and drop area into
+(`string`, `Element`, `Function`, or `UIPlugin`, default: `null`).
```
Murderlon, line 97: I think we always put `id` and `target` on top of the list

aduh95: Shouldn't we list the option in alphabetical order instead? That seems easier to enforce.

docs/user-interfaces/elements/drop-target.mdx
aduh95: We plan to have something at `/examples` don't we?

https://github.com/transloadit/uppy.io/blob/0f9e149b61682ae41579712e346cc98698970556/docusaurus.config.js#L58

Murderlon: No as discussed we will go live without it and potentially never have it as we emphasize our existing examples repo and codesandboxes. So at least for now we need to remove it.

docs/user-interfaces/elements/drop-target.mdx
Murderlon: It's a bit all over the place currently. I can make sorting part of #45. I can probably automate that with remark.

docs/user-interfaces/elements/drop-target.mdx
aduh95: > We should also create a codesandbox for this plugin and link it here.

I don't know what's needed for this. We should probably document some guidelines around it.

> No as discussed we will go live without it and potentially never have it as we emphasize our existing examples repo and codesandboxes.

There are 12 other pages linking to `/examples`, plus the main nav menu cited above. I don't think it makes sense to remove it from this PR. If we plan to remove them before going live, it can be done in a follow-up PR and doesn't have to block this one.

docs/user-interfaces/elements/drop-target.mdx
```diff
@@ -1 +1,103 @@
...
+
+[Try out the live example](/examples) or take it for a spin in
+[CodeSandbox](https://codesandbox.io/s/uppy-drag-drop-gyewzx?file=/src/index.js).
+
+:::
+
+## When should I use this?
+
+When you want to allow users to drag and drop files in your own UI, rather than
+in the [`Dashboard`](../dashboard) UI.
```
arturi, line 28:
```suggestion
in the [`Dashboard`](../dashboard) UI, or catch dropped files from anywhere on the page.
```

docs/user-interfaces/elements/drop-target.mdx
```diff
@@ -1 +1,103 @@
...
+uppy.use(DropTarget, {
+	target: document.body,
+});
+```
+
+## API
+
+### Options
+
+#### `onDragLeave(event)`
```
arturi, line 81: Do we want to leave the function name with arguments, so URLs won’t break if we change the arguments? https://transloadit.slack.com/archives/C0FMW9PSB/p1673972430908959

PR #67, opened 2023-02-15
add COMPANION_AWS_PREFIX

mifi:
https://github.com/transloadit/uppy/pull/4320

### DIFF
--- a/docs/companion.md
+++ b/docs/companion.md
@@ -457,6 +457,10 @@ The name of the bucket to store uploaded files in.
 
 The datacenter region where the target bucket is located.
 
+##### `s3.region` `COMPANION_AWS_PREFIX`
+
+An optional prefix for all uploaded keys.
+
 ##### `s3.awsClientOptions`
 
 You can supply any


docs/companion.md
```diff
@@ -457,6 +457,10 @@ The name of the bucket to store uploaded files in.
 
 The datacenter region where the target bucket is located.
 
+##### `s3.region` `COMPANION_AWS_PREFIX`
```
Murderlon, line 460: Is this correct? `region` and `prefix` sound like very different things?

mifi: whoops my bad

mifi, line 460:
```suggestion
##### `COMPANION_AWS_PREFIX`
```

docs/companion.md
```diff
@@ -457,6 +457,10 @@ The name of the bucket to store uploaded files in.
 
 The datacenter region where the target bucket is located.
 
+##### `COMPANION_AWS_PREFIX`
```
Murderlon, line 460: Is it correct that this is a standalone only feature? It's not possible to set a prefix when integrating into Express?

mifi: Ah yes, it's standalone only. But the same can be achieved with a custom `getKey` - let me add that.

docs/companion.md
```diff
@@ -457,6 +457,10 @@ The name of the bucket to store uploaded files in.
...
+
+An optional prefix for all uploaded keys.
```
mifi, line 462:
```suggestion
An optional prefix for all uploaded keys. This is a standalone-only option. The same can be achieved by the `getKey` option when using the express middleware.
```

mifi:
Prettier is pretty annoying. it doesn't give any feedback about what's wrong. could we auto-run `prettier --fix` in github actions for PRs? If not, I'm not sure how to use the `Suggest change` feature anymore now that we have prettier.

mifi:
Seems like I'm not the only one. there's an open issue about this at prettier github:
https://github.com/prettier/prettier/issues/6885

Murderlon (approved):

PR #69, opened 2023-03-14
Add the homepage

Murderlon:


src/pages/index.tsx
```diff
@@ -1,41 +1,371 @@
...
+  },
+  beforeDestroy () {
+    this.uppy.close({ reason: 'unmount' })
+  }
 }
+</script>
+`;
+
+const angularCode = `
+import { NgModule } from '@angular/core'
```
nqst, line 87:
```suggestion
const angularCode = `import { NgModule } from '@angular/core'
```

Would make the code start from line 1, currently it starts from line 2:

Screenshot 14-03-2023 at 15 12 37@2x

src/pages/index.tsx
```diff
@@ -1,41 +1,371 @@
...
-						to="/docs/intro"
-					>
-						Docusaurus Tutorial - 5min ⏱️
-					</Link>
-				</div>
-			</div>
-		</header>
-	);
+const dashboardCode = `
+import Uppy  from '@uppy/core'
```
nqst, line 27:
```suggestion
const dashboardCode = `import Uppy  from '@uppy/core'
```

src/pages/index.tsx
```diff
@@ -1,41 +1,371 @@
...
+
+const uppy = new Uppy().use(Webcam)
+
+function Component () {
+  return <Dashboard uppy={uppy} plugins={['Webcam']} />
+} 
+`;
+
+const vueCode = `
+<template>
```
nqst, line 56:
```suggestion
const vueCode = `<template>
```

src/pages/index.tsx
```diff
@@ -1,41 +1,371 @@
...
+const uppy = new Uppy()
+  .use(Dashboard, { target: '.DashboardContainer', inline: true })
+  .use(RemoteSources, { companionUrl: "http://companion.uppy.io" })
+  .use(Webcam, { target: Dashboard })
+  .use(ImageEditor, { target: Dashboard })
+  .use(Tus, { endpoint: 'https://tusd.tusdemo.net/files/' })
+`;
+
+const reactCode = `
+import React, { useEffect } from 'react'
```
nqst, line 43:
```suggestion
const reactCode = `import React, { useEffect } from 'react'
```

src/pages/index.tsx
```diff
@@ -1,41 +1,371 @@
...
+    UppyAngularDashboardModule,
+  ],
+  providers: [],
+  bootstrap: [AppComponent],
+})
+class {}
+`;
+
+const svelteCode = `
+<main>
```
nqst, line 108:
```suggestion
const svelteCode = `<main> 
```

src/pages/index.tsx
```diff
@@ -1,41 +1,371 @@
...
+					<div className={styles.dashboard}>
+						<BrowserOnly>
+							{() => {
+								const uppy = new Uppy().use(Webcam);
+								return (
+									<Dashboard uppy={uppy} height={300} plugins={['Webcam']} />
+								);
+							}}
+						</BrowserOnly>
+						<CodeBlock language="js" showLineNumbers>
```
nqst, line 224:
```suggestion
						<CodeBlock language="js">
```

Propose to remove `showLineNumbers` to save horizontal space — currently the code block has horizontal scrollbar even on desktop.

src/pages/index.tsx
```diff
@@ -1,41 +1,371 @@
...
+										onChange={(event) => setFramework(event.target.value)}
+									/>
+									<label htmlFor={name}>
+										<Icon />
+										<span>{name}</span>
+									</label>
+								</>
+							))}
+						</div>
+						<CodeBlock language="js" showLineNumbers>
```
nqst, line 304:
```suggestion
						<CodeBlock language="js">
```

src/pages/index.module.css
```diff
@@ -1,23 +1,394 @@
...
+.main section h2,
+.main section p {
+	max-width: 30rem;
+}
+
+.button {
+	color: var(--blue);
+	font-weight: 600;
+	background-color: var(--bg-blue);
+	padding: 0.5rem;
```
nqst, line 32:
```suggestion
	padding: 0.5rem 1rem;
```


Diff: 
Screenshot 14-03-2023 at 15 22 03

src/pages/index.module.css
```diff
@@ -1,23 +1,394 @@
...
-.heroBanner {
-	padding: 4rem 0;
+:root {
+	--pink: #f37;
+	--blue: #2563eb;
+	--bg-blue: #e0f2fe;
+}
+
+.main code {
+	font-size: 0.85rem !important;
```
nqst, line 10:
```suggestion
	font-size: 0.8rem !important;
```

Saving horizontal space

src/pages/index.tsx
```diff
@@ -1,41 +1,371 @@
...
+
+const providersIcons = [
+	'box.svg',
+	'unsplash.svg',
+	'googledrive.svg',
+	'dropbox.svg',
+	'instagram.svg',
+	'facebook.svg',
+	'onedrive.svg',
+];
```
nqst, line 133: Let's try the new colorful icons here? Believe it's going to look better, and we won't fail with following some brand guidelines.

docusaurus.config.js
```diff
@@ -71,29 +71,16 @@ const config = {
 			footer: {
```
nqst, line 71: Footer seems rather misaligned to me here. I propose not to center the copyright, at least on desktop.

image

src/pages/index.tsx
```diff
@@ -1,41 +1,371 @@
...
+							target="_blank"
+						>
+							Smashing Magazine
+						</a>
+					</figcaption>
+				</figure>
+			</header>
+
+			<main className={styles.main}>
+				<section className={styles['section-dashboard']}>
```
nqst, line 207: When the user's the screen real estate allows it, I propose to increase the width of this particular section — to fit the desktop version of Dashboard on larger screens. And we might also slightly increase the width of the code example block thanks to that.

nqst (changes requested):

src/pages/index.tsx
arturi: +1, I think those can safely be ignored on the homepage.

arturi:
Nice start!

I'd personally add more color/variety to the background, for starters to the hero, and make certain things wider, it feels a bit crampled to me. If we look at the default homepage:

image

We could take the full-width pink background and make the shooting Uppy white or another shade of pink.

I'd leave the margins in Much More section:

image

src/pages/index.tsx
arturi: :+1: I'd propose to make it full-width and code below or hidden in a tab. I think the centered full-width live example is more important than code as a homepage hero thingy.

src/pages/index.tsx
Murderlon: I still don't think it will look better and brand compliance is important for applications that need to go through approval. In our case that's the Uppy Dashboard, not our homepage.

nqst: Could we just try and compare tho?

src/pages/index.module.css
```diff
@@ -1,23 +1,405 @@
...
+	width: 150px;
+	display: none;
+}
+
+.upload svg {
+	transform: rotate(7deg) translateY(55px) translateX(-10px);
+	width: 100%;
+}
+
+.upload div {
```
arturi, line 151: I would add an extra css class vs targeting a div.

src/pages/index.module.css
```diff
@@ -1,23 +1,405 @@
...
+
+@media screen and (min-width: 50rem) {
+	.section-much-more ul {
+		display: grid;
+		grid-template-columns: repeat(2, minmax(0, 1fr));
+		gap: 0 1rem;
+		align-items: start;
+	}
+	.section-much-more li {
+		margin: 0;
```
arturi, line 376: It looks more readable without the margin, let’s remove?

src/pages/index.module.css
Murderlon: I'm not too worried here because `upload` are the flying uploads, they are specific and decorative, not reusable components. Also because nothing of the homepage CSS will be used anywhere but on this page.

src/pages/index.module.css
Murderlon: I think you mean _with_? Either way instead of margin I defined `gap` for the same effect. Does it need to be bigger?

src/pages/index.tsx
Murderlon: Resolved in Slack: keeping it as is.

PR #73, opened 2023-03-16
Add Locale docs

Murderlon:


arturi (changes requested):
I think we should at least copy the locale table as is https://uppy.io/docs/locales/#List-of-locale-packs. It might become outdated, but we can keep that in mind when new locale PRs come in, otherwise we might forget entirely.

PR #74, opened 2023-03-17
meta: show Prettier diff when prettier is invalid

aduh95:


mifi:
Nice! U want to add a comment to the original issue at prettier also or shall I?

mifi (approved):

mifi:
A future improvement could be  to add annotations in github actions so the error shows up directly:

https://github.com/orgs/community/discussions/26703

PR #76, opened 2023-03-20
Add locale switcher to example page

Murderlon:


src/pages/examples.tsx
```diff
@@ -230,22 +236,43 @@ export default function Examples() {
...
+								);
+							})}
+						</select>
 					</div>
 					<BrowserOnly>
 						{() => {
-							const uppy = new Uppy({ restrictions: state.restrictions })
+							const uppy = new Uppy({
+								restrictions: state.restrictions,
+								locale,
```
arturi, line 262: So it re-creates Uppy each time locale is updated? Fine in this case, but we could use `uppy.setOptions{ locale }`? Or we can’t because <BrowserOnly>?

arturi (approved):

src/pages/examples.tsx
Murderlon: Some experimentation could be done on this front. But in the current setup you can't because the controls, which would need access to `uppy` in scope, are outside the `BrowserOnly` component. We could make the whole page `BrowserOnly`, which for a client-side JS demo wouldn't be a bad thing, and then we should be able to make that work. The downside is that we would be fighting against React a bit, as the whole point is that you update state and React's render does the job (currently the case) but then we would be coming up with our own state updating cycle while working inside a React component.

PR #79, opened 2023-03-22
Add GitHub Action for deploying

Murderlon:
> **Note**
> One this is merged, it automatically does the deploy immediately. Before we merge, we must remove the uppy.io domain from `transloadit/uppy` and add it to this repository.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,0 +1,40 @@
+name: Deploy to GitHub Pages
+
+on:
+  push:
+    branches:
+      - main
+    # Review gh actions docs if you want to further define triggers, paths, etc
+    # https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#on
+
+jobs:
+  deploy:
+    name: Deploy to GitHub Pages
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 18
+          cache: yarn
+
+      - name: Install dependencies
+        run: yarn install --frozen-lockfile
+      - name: Build website
+        run: yarn build
+
+      # Popular action to deploy to GitHub Pages:
+      # Docs: https://github.com/peaceiris/actions-gh-pages#%EF%B8%8F-docusaurus
+      - name: Deploy to GitHub Pages
+        uses: peaceiris/actions-gh-pages@v3
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          # Build output to publish to the `gh-pages` branch:
+          publish_dir: ./build
+          # The following lines assign commit authorship to the official
+          # GH-Actions bot for deploys to `gh-pages` branch:
+          # https://github.com/actions/checkout/issues/13#issuecomment-724415212
+          # The GH actions bot is used by default if you didn't specify the two fields.
+          # You can swap them out with your own user credentials.
+          user_name: github-actions[bot]
+          user_email: 41898282+github-actions[bot]@users.noreply.github.com


.github/workflows/deploy.yml
```diff
@@ -0,0 +1,40 @@
...
+
+      - name: Install dependencies
+        run: yarn install --frozen-lockfile
+      - name: Build website
+        run: yarn build
+
+      # Popular action to deploy to GitHub Pages:
+      # Docs: https://github.com/peaceiris/actions-gh-pages#%EF%B8%8F-docusaurus
+      - name: Deploy to GitHub Pages
+        uses: peaceiris/actions-gh-pages@v3
```
aduh95, line 29: Can we not use an external action for that? AFAICT we just have to push to the gh-pages branch

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,40 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 18
+          cache: yarn
+
+      - name: Install dependencies
+        run: yarn install --frozen-lockfile
```
aduh95, line 22: `--frozen-lockfile` is for Yarn 1.x
```suggestion
        run: yarn install --immutable
```

aduh95:
> Before we merge, we must remove the uppy.io domain from `transloadit/uppy` and add it to this repository.

Why? There's no reason to do one before the other, I would argue that, on the contrary, we should first confirm it works before doing the switch.

.github/workflows/deploy.yml
Murderlon: This is the recommended way for docusaurus according to their docs. It's also a well maintained.

https://github.com/peaceiris/actions-gh-pages

https://docusaurus.io/docs/deployment#triggering-deployment-with-github-actions

Murderlon:
> Why? There's no reason to do one before the other, I would argue that, on the contrary, we should first confirm it works before doing the switch.

True, we don't have to switch the domain immediately. Perhaps the `CNAME` record needs to be removed from the repository then.

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,40 @@
+name: Deploy to GitHub Pages
+
+on:
+  push:
+    branches:
+      - main
+    # Review gh actions docs if you want to further define triggers, paths, etc
+    # https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#on
+
```
aduh95, line 9: Let's restrict the permissions to what's strictly needed so we don't have to worry as much about using external workflows.
```suggestion

permissions:
  contents: write

```

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,40 @@
...
+          node-version: 18
+          cache: yarn
+
+      - name: Install dependencies
+        run: yarn install --immutable
+      - name: Build website
+        run: yarn build
+
+      # Popular action to deploy to GitHub Pages:
+      # Docs: https://github.com/peaceiris/actions-gh-pages#%EF%B8%8F-docusaurus
```
aduh95, line 27:
```suggestion
      # The following actions is Docosorus recommended way to publish to GH Pages.
```

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,43 @@
...
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          # Build output to publish to the `gh-pages` branch:
+          publish_dir: ./build
+          # The following lines assign commit authorship to the official
+          # GH-Actions bot for deploys to `gh-pages` branch:
+          # https://github.com/actions/checkout/issues/13#issuecomment-724415212
+          # The GH actions bot is used by default if you didn't specify the two fields.
+          # You can swap them out with your own user credentials.
+          user_name: github-actions[bot]
+          user_email: 41898282+github-actions[bot]@users.noreply.github.com
```
aduh95, line 43: Since it says it's the default, maybe we should just not bother with that?
```suggestion
```

aduh95 (approved):

PR #83, opened 2023-03-23
Fix vue doc.

mathieutu:
Hey,
Thanks for the work!

The vue doc has been pasted from React one, but the example syntax is wrong 🙂.


aduh95 (approved):
Ah good catch! Thanks for the fix.

PR #87, opened 2023-03-27
CDN version: use package.json from uppy

arturi:


Murderlon (commented):
This is way simpler than a cross-repo GH Action, makes sense 👍 

Two things:
- We should make `uppy` a dev dependency as we only use it at build time
- We need to configure dependabot to immediately make a PR for any version bump of `uppy`.

aduh95:
I don't have access to the repo settings for Dependabot. Note that we might want to have some automation anyway as we want to merge the doc PR as soon as a new Uppy version is released, so it might make sense to build something custom.

aduh95:
> * We need to configure dependabot to immediately make a PR for any version bump of `uppy`.

This can be considered done now that we have the `update-deps` action I think.

Murderlon:
@arturi just the dev dep comment then 👍

Murderlon (approved):

PR #91, opened 2023-03-29
Add Golden Retriever to examples page

mifi:
i struggle with getting the types right, maybe someone could help

### DIFF
--- a/src/pages/examples.module.css
+++ b/src/pages/examples.module.css
@@ -63,6 +63,7 @@
 	max-width: 30rem;
 	border-radius: 5px;
 	padding: 1rem;
+	font-size: 0.8rem;
 }
 
 .options-wrapper > div {


Murderlon (commented):
Thanks for adding!

Can you also make the `initialState.theme` `light`? It's a bug currently.

src/pages/examples.tsx
```diff
@@ -90,6 +99,9 @@ function reducer(state: State, action: Action) {
 				...state,
 				plugins: state.plugins.filter((p) => p !== action.value),
 			};
+		case 'enableGoldenRetriever': {
```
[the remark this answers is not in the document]
mifi: I thought plugins are for the UI plugins. I tried to put goldenretriever inside the plugins array but it crashes. I think because goldenretriever is not a ui plugin and shouldnt be sent to <Dashboard plugins

src/pages/examples.tsx
```diff
@@ -166,11 +178,64 @@ const options = [
...
-		options: [{ label: 'Restrictions', type: 'restrictions' }],
+		options: [
+			{ label: 'Restrictions', type: 'restrictions' },
+			{ label: 'Golden Retriever', type: 'enableGoldenRetriever' },
+		],
 	},
 ];
 
-const companionUrl = 'https://companion.uppy.io';
+const Uppy = memo(({ state, locale }) => {
```
mifi, line 188:
```suggestion
const Uppy = (({ state, locale }) => {
```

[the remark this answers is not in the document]
mifi: Ok i removed memo

src/pages/examples.tsx
```diff
@@ -166,11 +178,64 @@ const options = [
...
+		if (state.enableGoldenRetriever) {
+			ret.use(GoldenRetriever);
+		}
+
+		return ret;
+	}, [state, locale]);
+
+	const [uppy, setUppy] = useState(createUppy);
+
+	useEffect(() => setUppy(createUppy()), [createUppy]);
```
Murderlon, line 216: I think the result of using state + effect isn't any different from initliazing uppy in the render? If we use state, we should use `setOptions` in a effect, but since we're not doing that, we can keep it simple?

mifi: I didn't want to implement setOptions in useEffect because as we've seen in previous discussions, it's quite complicated to implement it correctly. But using a useEffect and definining dependencies like this ensures that Uppy doesn't get accidentally reset if something causes a re-render (e.g. if something elsewhere in the app gets re-rendered)

mifi:
btw I think typescript checks are not working. because i had a typescript error before pushing

PR #94, opened 2023-03-29
experiment with Merge Queues

aduh95:
Merge Queues will run the checks on the commit that's about to merge to target branch, and will abort if a failure is detected. The difference with auto-merges is that the checks are re-run just before merging, while auto-merges behaves as if the target branch have not changed since the checks was last run.


Murderlon:
Something seems off, both checks passed but it's not being merged.

Murderlon (approved):

PR #98, opened 2023-04-03
meta: add VSCode extensions recommendations

aduh95:


aduh95:
@kvz Should we do that for the other repos that are using Prettier?

PR #102, opened 2023-04-05
Debug log, https, expose uppy on window

arturi:


src/pages/examples.tsx
```diff
@@ -274,6 +276,9 @@ export default function Examples() {
 								.use(Unsplash, { companionUrl })
 								.use(Box, { companionUrl });
 
+							// Expose for easier debugging
+							globalThis.uppy = uppy;
```
Murderlon, line 280: You need to place this in a `useEffect` because you only want this to happen once, not every state change.

src/pages/index.tsx
```diff
@@ -221,21 +224,22 @@ export default function Home(): JSX.Element {
...
+									.use(GoogleDrive, { companionUrl })
+									.use(Dropbox, { companionUrl })
+									.use(Instagram, { companionUrl })
+									.use(Url, { companionUrl })
+									.use(OneDrive, { companionUrl })
+									.use(Unsplash, { companionUrl })
+									.use(Box, { companionUrl });
+
+								// Expose for easier debugging
+								globalThis.uppy = uppy;
```
Murderlon, line 242: same here

src/pages/examples.tsx
mifi: But Uppy also gets recreated on every state change, so an effect wouldn’t make a difference. I think as long as uppy is recreated on every render, this is the correct way to do it.

src/pages/index.tsx
```diff
@@ -43,7 +46,7 @@ import Tus from '@uppy/tus'
 
 const uppy = new Uppy()
   .use(Dashboard, { target: '.DashboardContainer', inline: true })
-  .use(RemoteSources, { companionUrl: "http://companion.uppy.io" })
+  .use(RemoteSources, { companionUrl: "https://companion.uppy.io" })
```
mifi, line 49: Shouldn’t this use the above companionUrl variable?

arturi: It's a string, so wanted to be more explicit, but sure, changed.

PR #103, opened 2023-04-05
meta: add action to update Uppy packages

aduh95:


.github/workflows/update-deps.yml
```diff
@@ -0,0 +1,63 @@
+name: Tools and deps update
+on:
+  schedule:
+    # Run once a week at 00:05 AM UTC on Sunday.
+    - cron: 5 0 * * 0
```
Murderlon, line 5: Do we need a schedule?

.github/workflows/update-deps.yml
```diff
@@ -0,0 +1,63 @@
...
+
+  workflow_dispatch:
+  pull_request:
+    types: [opened, synchronize, reopened]
+    paths:
+      - .github/workflows/update-deps.yml
```
Murderlon, line 11: Maybe I misunderstand but don't we want to do this only before a deploy on `main`?

.github/workflows/update-deps.yml
aduh95: It should avoid us forgetting to run it manually.

.github/workflows/update-deps.yml
aduh95: Hum I think we only want it after each Uppy release.

Murderlon: But now it runs on every PR?

mifi: maybe if we could have uppy publish to npm (uppy@latest) on every commit, then we could use that here so we could update uppy.io to uppy@latest using the workflow_dispatch whenever we need that (or the schedule will)

aduh95: > But now it runs on every PR?

It runs on every PRs that modify `.github/workflows/update-deps.yml`.

Murderlon (approved):

PR #115, opened 2023-05-09
fix companion example

mifi:
closes https://github.com/transloadit/uppy/issues/4441

### DIFF
--- a/docs/companion.md
+++ b/docs/companion.md
@@ -130,7 +130,8 @@ const options = {
 	filePath: '/path/to/folder/',
 };
 
-app.use('/companion', companion.app(options));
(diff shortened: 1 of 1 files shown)


docs/companion.md
```diff
@@ -130,7 +130,8 @@ const options = {
 	filePath: '/path/to/folder/',
 };
 
-app.use('/companion', companion.app(options));
+const { app: companionApp } = companion.app(companionOptions)
+app.use(companionApp)
```
aduh95, line 134:
```suggestion
const { app: companionApp } = companion.app(companionOptions);
app.use(companionApp);
```

PR #118, opened 2023-05-17
Consistent typography

Murderlon:
Make the typography consistent again. I naively made this website only for desktop sizes which are around my laptop size, but in screenshares I've seen that on @arturi's laptop the typography is too big. Artur later implemented some fixes to make things a bit smaller here and there, but this became a bit inconsistent in my opinion. Instead, everything is now base font size but between 1000 and 1300px laptop screens the base font size is decreased, making it overall more fitting.

I also updated the blog post footer links:

 
Screenshot 2023-05-17 at 12 27 42

### DIFF
--- a/src/css/custom.css
+++ b/src/css/custom.css
@@ -92,6 +92,12 @@
 	--ifm-avatar-intro-margin: 0.5rem;
 }
 
+@media screen and (min-width: 1000px) and (max-width: 1300px) {
+	:root {
+		--ifm-font-size-base: 16px;
+	}
+}
+
 h1 {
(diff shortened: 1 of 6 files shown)


Murderlon:
@arturi would be nice to merge this before announcing the new website :)

arturi:
This is how examples page is rendered on my screen:

image

arturi:
This footer change is intentional also, right, you suggest no button, but a link? Sounds good, not sure if it's worth duplicating the Github we have in the top menu

image

arturi:
Examples needs to be fixed — more space to the settings column. Otherwise looks good to me!

PR #126, opened 2023-06-10
Fix authProvider doc

mifi:
make it static

See https://github.com/transloadit/uppy/issues/4460


Murderlon (approved):

Murderlon:
Linter is failing though

mifi:
ah yes prettier, will fix

PR #133, opened 2023-06-29
@uppy/aws-s3-multipart: add docs for signing on the client

aduh95:
Refs: https://github.com/transloadit/uppy/pull/4519

### DIFF
--- a/docs/uploader/aws-s3-multipart.mdx
+++ b/docs/uploader/aws-s3-multipart.mdx
@@ -430,6 +430,7 @@ sort of caching process to avoid requesting more temporary token than necessary.
     - `AccessKeyId`: `string`
     - `SecretAccessKey`: `string`
     - `SessionToken`: `string`
+    - `Expiration`: `string`
   - `bucket`: `string`
   - `region`: `string`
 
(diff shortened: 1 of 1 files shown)


docs/uploader/aws-s3-multipart.mdx
```diff
@@ -410,8 +410,83 @@ uppy.use(AwsS3Multipart, {
 });
 ```
 
+#### `getTemporarySecurityCredentials(options)`
+
+:::note
+
+When using [Companion][companion docs] as a backend, you can pass `true` instead
```
Murderlon, line 417:
```suggestion
If you're already using [Companion][companion docs] for remote files, you can pass `true.
It is not recommended to introduce Companion just for S3 signing, use your own backend instead.
```

docs/uploader/aws-s3-multipart.mdx
```diff
@@ -410,8 +410,83 @@ uppy.use(AwsS3Multipart, {
 });
 ```
 
+#### `getTemporarySecurityCredentials(options)`
```
Murderlon, line 413: perhaps an extra `:::note` that links to the example in the uppy repo?

docs/uploader/aws-s3-multipart.mdx
aduh95: I've copied the wording from https://github.com/transloadit/uppy.io/blob/d8ac350c76175be0df47e009afdf1d82a47dc455/docs/uploader/aws-s3.mdx#L322-L323
I get what you mean, but I think the wording in your suggestion can be a bit confusing, I'd rather stick to the current wording.

docs/uploader/aws-s3-multipart.mdx
aduh95: Currently there are no examples that uses that option, so I guess this should wait until there's one (which cannot happen until that option ships into a release, because we are using the CDN in our demo)

docs/uploader/aws-s3-multipart.mdx
Murderlon: You referenced example is about not defining this option, for which I agree. But this is passing a boolean instead of implementing a function yourself. I think it's still good to reiterate that you shouldn't introduce Companion to make it 'easier'.

arturi (approved):

Murderlon (commented):
I think we are missing a "why". 

- Add a small section in "When should I use this"? to make the user aware this exists (instead of accidentally stumbling upon this here) and when/why they want this over signing URLs.
- Add the why below the option too

Murderlon:
I expected a bit more effort in explaining this to the user. Imagine coming to this plugin, you either completely miss this option or you coincidentally find it but you have no idea what it is. 

This was the text:
>  A boolean, or an (async) function that returns a boolean which is called each time a signed URL needs to be generated.

- How is signed URL different, thought we were already signing URLs?
- What is this?
- When do I want this?
- What is the security trade off?

Tried to address it in 34558734ce8f21ecc8a8a72bb829bef3b0e92f38. 

I wrote "this is a security trade-off" but I think it needs to be more explicit. @aduh95 can you rewrite this sentence to indicate when you do _not_ want to use this?

docs/uploader/aws-s3-multipart.mdx
```diff
@@ -27,17 +27,25 @@ _client-to-server-to-storage_ (such as [Transloadit](/docs/transloadit) or
...
+example, if you are uploading files that are only a couple kilobytes with a
+100ms roundtrip latency, you are spending 400ms on overhead and only a few
+milliseconds on uploading.
+
+**In short**
+- If you are uploading large files (100&nbsp;MB+), we recommend
 `@uppy/aws-s3-multipart`, otherwise [`@uppy/aws-s3`][].
+- If you prefer to have less overhead (+20% upload speed)
+  you can use temporary S3 credentials with
+  [`getTemporary`](#gettemporarysecuritycredentialsoptions).
```
aduh95, line 45:
```suggestion
  [`getTemporarySecurityCredentials `](#gettemporarysecuritycredentialsoptions).
```

Murderlon:
imo there is no excuse to create a poorly documented feature. A lot of time went into researching and creating this, let that work be paid off by informing our users when this great feature is great, and when it isn't.

aduh95:
> * What is the security trade off?

I couldn't answer this question myself, and I'm not sure there's an answer that can apply for the general case, I feel it should be a case-by-case basis.

Murderlon (approved):
LGTM, thanks

PR #143, opened 2023-07-17
chore: fix auto-update workflow

aduh95:
Not sure why I went with `@uppy/core` originally, using the Uppy version seems much more logical.


Murderlon (approved):
Ah yes

PR #153, opened 2023-08-07
Document how to allow duplicate files

Murderlon:
Ref: https://github.com/transloadit/uppy/pull/4594

### DIFF
--- a/docs/uppy-core.mdx
+++ b/docs/uppy-core.mdx
@@ -316,10 +316,11 @@ Metadata can also be added from a `<form>` element on your page, through the
 #### `onBeforeFileAdded(file, files)`
 
 A function called before a file is added to Uppy (`Function`, default:
-`(file) => file`).
+`(files, file) => !Object.hasOwn(files, file.id)`).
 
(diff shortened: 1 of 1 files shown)


aduh95 (approved):

docs/uppy-core.mdx
```diff
@@ -345,6 +346,16 @@ purposes via [`uppy.log()`](#logmessage-type).
 <details>
   <summary>Filter, change, and abort example</summary>
 
+Allow all files, also duplicate files. This will override the file if it has not
```
arturi, line 349:
```suggestion
Allow all files, also duplicate files. This will overwrite the file (clear meta data, reset image editor modifications) if it has not
```

arturi (approved):

docs/uppy-core.mdx
Murderlon: How about this? because replace means _everything_ you could possibly do with uppy files.

```suggestion
Allow all files, also duplicate files. This will replace the file if it has not
```

PR #155, opened 2023-08-10
More compact Table of Contents

nqst:
I noticed that the Table of Contents is overly detailed in the recent blog post. It even causes an extra scrollbar:

Screenshot 2023-08-10 at 13 34 34@2x

I propose that we only show the `<h2>` tags here, as I don't think there's a need to provide such a complete structure. Here's how the more compact version looks: 

Screenshot 2023-08-10 at 13 38 52@2x

* * *

**Note:** This change also affects Docs. Also LGTM there: 

Screenshot 2023-08-10 at 13 40 55

**Note 2**: I also slightly reduced the post headings font size. You can notice the difference on the first two screenshots.

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -156,6 +156,9 @@ const config = {
 				theme: lightCodeTheme,
 				darkTheme: darkCodeTheme,
 			},
+			tableOfContents: {
+				maxHeadingLevel: 2,
+			}
 		}),
 };
 
--- a/src/css/custom.css
+++ b/src/css/custom.css
@@ -117,7 +117,7 @@ p > code {
 	font-size: var(--ifm-font-size-base);
 }
 
-/* 
+/*
 * Admonitions have long generated class names.
 * We want to change the heading font, but it's not an config option.
 */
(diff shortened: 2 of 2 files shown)


arturi:
Thanks for the proposal! I think I agree about the blog, but not so sure about the docs. It might help to have everything in the sidebar so you can navigate to an option or method quickly.

@Murderlon what do you think?

Murderlon (changes requested):
We should not decrease the heading level of the table of contents, in my opinion this is essential in navigating in all places. I'm okay with alternative styling solutions.

arturi:
@Murderlon You think it's important for the blog posts? I'd argue we can hide the sidebar for them completely. (https://deno.com/blog/node-config-hell, https://transloadit.com/blog/2023/07/wasabi-robots/)

Murderlon:
I mean since there is no option to disable it only on the blog page, I don't think we should set the setting, but I am open to alternative solutions with CSS only on the blog page.

nqst:
> I mean since there is no option to disable it only on the blog page

Actually, we can do it. I couldn't found a global setting which would affect all blog posts, but we can do something in a post's YAML front matter: either set `hide_table_of_contents` to `true`, or specify the `toc_max_heading_level` there.

I tried both, and ended up with setting `toc_max_heading_level: 2` for that particular post.

Murderlon (approved):
Great, thanks!

arturi:
Ok, but with this proposed change, we'll have to set this option manually for all blog posts? Also couldn't find a way to set it to blog-only in Docusaurus.

nqst:
> we'll have to set this option manually for all blog posts? 

I don't think so. That particular post had such a long ToC so I wanted to hide it. I think for the remaining majority of posts having it enabled (= no need to set any options) should be fine.

PR #156, opened 2023-08-16
Fix broken link in docs

choptastic:
The link in the warning on https://uppy.io/docs/aws-s3/ is broken and leads to a 404. This fixes that.


Murderlon (approved):
Thanks

PR #166, opened 2023-09-14
Update companion.md

mifi:
improve descriptiuon of middleware/standalone. I think standalone should be the first choice

### DIFF
--- a/docs/companion.md
+++ b/docs/companion.md
@@ -71,10 +71,9 @@ files to arrive at Transloadit servers, much like Uppy.
 Companion is installed from npm. Depending on how you want to run Companion, the
 install process is slightly different. Companion can be integrated as middleware
 into your [Express](https://expressjs.com/) app or as a standalone server.
-
-```bash
-npm install @uppy/companion
-```
+Most people probably want to run it as a standalone server, while the middleware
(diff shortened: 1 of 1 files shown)


docs/companion.md
```diff
@@ -86,7 +85,56 @@ accept improvements in this area, but we can’t provide support.
 
 :::
 
-### Express middleware
+### Standalone mode
+
+You can use the standalone version if you want to run Companion as it’s own Node
```
aduh95, line 90:
```suggestion
You can use the standalone version if you want to run Companion as it’s own Node.js
```

aduh95 (commented):
There are some features who are available only when using Companion as a middleware (e.g. dynamic buckets), so I'm not so sure the standalone mode should be put too much in front. wdyt?

mifi:
Another reason why I want to place standalone first is because we get a lot of support requests from people who try to set it up using a middlware, but have problems because they don't setup their express server correctly (like in standalone mode). If we encourage standalone first, then once they get it working with standalone, they can more easily move on to the (more advanced) middleware mode hopefully without needing our help.

Murderlon:
I'm fine with the order change. CI is failing though.

mifi:
fixed!

PR #167, opened 2023-09-20
Document the redirect URL when using Transloadit

Murderlon:


cdr-chakotay:
I think this helps a lot already :)

mifi (approved):
🙌

PR #171, opened 2023-10-04
fix navbar going behind uppy

mifi:
on the examples page. note: this is a workaround and we might want to instead fix it by not using z-indexes inside uppy, but that's probably a bigger task


Murderlon (approved):

arturi:
Yes, thank you! It's on my list for Uppy, but indeed should have fixed the website right away.

Issue #123, opened 2023-05-31
Document the range of storage for Golden Retriever


Murderlon:
Document how much a service worker can take up:

This might help:
https://love2dev.com/blog/what-is-the-service-worker-cache-storage-limit/

PR #173, opened 2023-10-05
Document range of storage for Golden Retriever

Murderlon:
Closes https://github.com/transloadit/uppy.io/issues/123

### DIFF
--- a/docs/golden-retriever.mdx
+++ b/docs/golden-retriever.mdx
@@ -32,7 +32,9 @@ If restore is unsuccessful for certain files, they will be marked as â€œghostsâ€
 the Dashboard UI, and a message + button offering to re-select those files will
 be displayed.
 
-Checkout the [storage quotas](https://developer.mozilla.org/en-US/docs/Web/API/Storage_API/Storage_quotas_and_eviction_criteria#other_web_technologies) on MDN to see how much data can be stored depending on the device and browser.
+Checkout the
+[storage quotas](https://developer.mozilla.org/en-US/docs/Web/API/Storage_API/Storage_quotas_and_eviction_criteria#other_web_technologies)
+on MDN to see how much data can be stored depending on the device and browser.
 
 ## When should I use this?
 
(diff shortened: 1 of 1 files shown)


docs/golden-retriever.mdx
```diff
@@ -32,6 +32,21 @@ If restore is unsuccessful for certain files, they will be marked as “ghosts
...
+
+The amount of storage is dependent upon how much available disk space the device
+has. The general range is between 50MB and 20GB.
+
+| Disk Space | Storage Limit                     |
+| ---------- | --------------------------------- |
+| < 8GB      | 50MB                              |
+| > 8-32GB   | 500MB                             |
+| > 32-128GB | 4% of volume size                 |
+| > 128GB    | 4% or 20GB (whichever is smaller) |
```
arturi, line 46: Should we link to MDN or some other source, in case this changes for instance?

Murderlon: There is no single source of truth link, the amount available to service worker caching isn't defined in the specification. The amount varies by browser depending on device and storage conditions. This is a generalised version of the differences and it unlikely to change.

arturi: Found this, wouldn't hurt? https://developer.mozilla.org/en-US/docs/Web/API/Storage_API/Storage_quotas_and_eviction_criteria

Murderlon: Wow I missed that one, that's actually quite good

docs/golden-retriever.mdx
```diff
@@ -32,6 +32,21 @@ If restore is unsuccessful for certain files, they will be marked as “ghosts
...
+
+</details>
```
Murderlon, line 48:
```suggestion
Checkout the [storage quotas](https://developer.mozilla.org/en-US/docs/Web/API/Storage_API/Storage_quotas_and_eviction_criteria#other_web_technologies) on MDN to see how much data can be stored depending on the device and browser.
```

arturi (approved):

PR #176, opened 2023-10-19
Improve React docs

Murderlon:
@nickrttn if there was anything confusing in your journey to add Uppy to Transloadit that is not documented here please point it out as well.


docs/framework-integrations/react.mdx
```diff
@@ -42,62 +42,131 @@ You also need to install the UI plugin you want to use. For instance,
...
+
+const files = useUppyState(uppy, (state) => state.files);
+const totalProgress = useUppyState(uppy, (state) => state.totalProgress);
+// We can also get specific plugin state.
+// Note that the value on `plugins` depends on the `id` of the plugin.
+const metaFields = useUppyState(
+	uppy,
+	(state) => state.plugins?.Dashboard?.metaFields,
+);
+```
```
nickrttn, line 83: This can be abstracted one level deeper to offer a better DX, while avoiding prop drilling with the `uppy` variable:

```jsx
const UppyContext = createContext(null)

function useUppyState(selector) {
  const uppy = useContext(UppyContext)
  if (!uppy) throw new Error('needs to be in context')
  return whateverUseUppyStateDidBefore(uppy, selector)
}

function useUppy() {
  return useContext(UppyContext)
}

function UppyProvider(props) {
  const uppy = useRef()

  if (!uppy.current) {
    uppy.current = new Uppy(props.options)
  }

  return (
    <UppyContext.Provider value={uppy.current}>
      {props.children}
    </UppyContext.Provider>
  )
}
```

This would allow folks:

- to wrap each component that needs its own instance of Uppy with the provider
- to call `useUppyState` anywhere in the tree without having to pass `uppy` down
- to update Uppy's options in `useEffect` hooks or event handlers from inside the tree by using `useUppy`

Murderlon: I'll copy my comment from the PR itself as I still stand by it.

---

I thought about it but refrained from doing so:

> Context is very tempting to use! However, this also means it’s too easy to overuse it. **Just because you need to pass some props several levels deep doesn’t mean you should put that information into context** — [React docs](https://react.dev/learn/passing-data-deeply-with-context#before-you-use-context)

The React docs use cases for adding context also don't align with what Uppy is used for. Realistically, there is only one page where people upload in most apps and it would be overkill to provide a context for this in my opinion. I'd rather keep it simple and explicit for now. Perhaps reconsidering when we see people needing this more.

docs/framework-integrations/react.mdx
```diff
@@ -42,62 +42,131 @@ You also need to install the UI plugin you want to use. For instance,
...
-- You have to pass the IDs of your `use`d plugins to the `plugins` array prop so
-  Dashboard knows it needs to render them.
-- An Uppy instance can’t be used by more than one component. Make sure you are
-  using a unique Uppy instance per component.
+Use this hook when you need to access Uppy’s state reactively. Most of the
+times, this is needed if you are building a custom UI for Uppy in React.
 
-:::
+```js
+// IMPORTANT: passing an initializer function to prevent the state from recreating.
```
mifi, line 72:
```suggestion
// IMPORTANT: passing an initializer function to prevent Uppy from being reinstantiated on every render.
```

docs/framework-integrations/react.mdx
```diff
@@ -42,62 +42,131 @@ You also need to install the UI plugin you want to use. For instance,
...
-// Don't forget the CSS: core and the UI components + plugins you are using.
 import '@uppy/core/dist/style.min.css';
 import '@uppy/dashboard/dist/style.min.css';
 import '@uppy/webcam/dist/style.min.css';
 
-// Don’t forget to keep the Uppy instance outside of your component.
-const uppy = new Uppy().use(Webcam);
-
 function Component() {
+	// IMPORTANT: passing an initializer function to prevent the state from recreating.
```
mifi, line 117:
```suggestion
	// IMPORTANT: passing an initializer function to prevent Uppy from being reinstantiated on every render.
```

docs/framework-integrations/react.mdx
```diff
@@ -42,62 +42,131 @@ You also need to install the UI plugin you want to use. For instance,
...
+		</>
+	);
+}
+
+export default function App() {
+	// keeping the uppy instance alive above the pages the user can switch during uploading
+	const [uppy] = useState(() => new Uppy());
+
+	return (
+		<>
```
mifi, line 154:
```suggestion
                // Add your router here
		<>
```

mifi (approved):

PR #179, opened 2023-10-24
Document S3 endpoint for Companion

### LINKED ISSUE #178: @uppy/companion: support non-aws S3 endpoints via environment variables
### Initial checklist

- [X] I understand this is a feature request and questions should be posted in the [Community Forum](https://community.transloadit.com/)
- [X] I searched [issues](https://github.com/transloadit/uppy/issues?q=is%3Aissue) and couldn’t find anything (or linked relevant results below)

I am aware of a custom solution like here: 
https://github.com/transloadit/uppy/blob/main/examples/digitalocean-spaces/server.cjs#L39
but prefer not to write my own server files etc. if possible - just to specify a different s3 endpoint.

### Problem

I am using a docker-compose file to host the transloadit/companion image. 
I found options to declare AWS S3 secret, key etc. but not to specify the s3 endpoint in the .env file / environment. 

### Solution

I'd like to be able to define a `COMPANION_S3_ENDPOINT` in the .env file or environment settings to use other s3 compatible storage providers.

### Alternatives

config files or other settings. I'd prefer this configuration to be as simple as possible. If I overlooked a similar setting - sorry! I'm happy to hear about that.

Murderlon:
Closes #178

### DIFF
--- a/docs/companion.md
+++ b/docs/companion.md
@@ -463,6 +463,10 @@ which has only the secret, nothing else.
 
 :::
 
+##### `s3.endpoint` `COMPANION_AWS_ENDPOINT`
+
+URL to the S3 (compatible) service.
+
 ##### `s3.bucket` `COMPANION_AWS_BUCKET`
 
 The name of the bucket to store uploaded files in.
--- a/docs/companion.md
+++ b/docs/companion.md
@@ -465,8 +465,8 @@ which has only the secret, nothing else.
 
 ##### `s3.endpoint` `COMPANION_AWS_ENDPOINT`
 
-Optional URL to a custom S3 (compatible) service.
-Otherwise uses the default from the AWS SDK.
+Optional URL to a custom S3 (compatible) service. Otherwise uses the default
+from the AWS SDK.
 
 ##### `s3.bucket` `COMPANION_AWS_BUCKET`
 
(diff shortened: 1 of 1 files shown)


docs/companion.md
```diff
@@ -463,6 +463,10 @@ which has only the secret, nothing else.
 
 :::
 
+##### `s3.endpoint` `COMPANION_AWS_ENDPOINT`
+
+URL to the S3 (compatible) service.
```
mifi, line 468:
```suggestion
URL to a custom S3 (compatible) service.
```

Murderlon: custom implies that it's optional? is that the case? Does it have a default?

mifi: I believe the default lies somewhere inside aws sdk. And it’s probably amazon-aws-something.com

docs/companion.md
```diff
@@ -463,6 +463,10 @@ which has only the secret, nothing else.
 
 :::
 
+##### `s3.endpoint` `COMPANION_AWS_ENDPOINT`
+
+Optional URL to a custom S3 (compatible) service. Otherwise uses the default from the AWS SDK.
```
Murderlon, line 468:
```suggestion
Optional URL to a custom S3 (compatible) service.
Otherwise uses the default from the AWS SDK.
```

PR #191, opened 2023-11-30
Use prettier spaces intead of tabs to fix code blocks

Murderlon:


aduh95:
What does it mean "to fix code blocks"?

Murderlon:
To not have enormous spacing:

Screenshot 2023-11-30 at 14 14 27

Unfortunately no Docusaurus option for this to solving it with spaces.

aduh95:
Wouldn't https://caniuse.com/css3-tabsize be the thing we need?

Murderlon:
Just tested it on the website and that works too. Maybe better to try that indeed.

PR #192, opened 2023-11-30
Set tab-size to 2

Murderlon:
Supersedes #191

### DIFF
--- a/src/css/custom.css
+++ b/src/css/custom.css
@@ -98,6 +98,10 @@
 	}
 }
 
+body {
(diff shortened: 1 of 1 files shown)


src/css/custom.css
```diff
@@ -98,6 +98,10 @@
 	}
 }
 
+body {
+	tab-size: 2;
+}
```
aduh95, line 103: Should we move this to `:root` above?

aduh95 (approved):

PR #197, opened 2023-12-12
update Plausible metadata

aduh95:
We are no longer going to use the self-hosted instance.


arturi (approved):

arturi:
We need to update https://uppy.io/privacy-policy/ as well, it says “self-hosted”.

PR #200, opened 2023-12-28
Fix SEO issues to impove the Ahrefs' Health Score

nqst:
Made some changes that should significantly improve our SEO [Health Score](https://help.ahrefs.com/en/articles/1424673-what-is-health-score-and-how-is-it-calculated-in-ahrefs-site-audit) provided by Ahrefs (it's 73% as of Dec 2023):

- Fixed a lot of broken links.
- Fixed that some links that were incorrectly parsed by Docusaurus by using a `pathname://` workaround (see [/docusaurus/issues/6282](https://github.com/facebook/docusaurus/issues/6282)). Not super happy with that solution, but I couldn't find a better one.
- Fixed that Open graph URLs did not match Canonical URLs.
- Fixed incorrect Twitter card formatting.
- Significantly optimized some heavyweight images.

**Remaining issues:**
- We're using [`plugin-client-redirects`](https://docusaurus.io/docs/api/plugins/@docusaurus/plugin-client-redirects) for Docusaurus but it's bad for SEO. Ideally, we should use server-side redirects instead. Is it possible?


docs/uploader/aws-s3-multipart.mdx
```diff
@@ -191,7 +191,7 @@ const uppy = new Uppy()
 :::warning
 
 Until the next major version, not setting this option uses the
-[legacy version of this plugin](./aws-s3). This is a suboptimal experience for
+[legacy version of this plugin](../aws-s3). This is a suboptimal experience for
```
aduh95, line 194: This change looks odd, `aws-s3.mdx` is a sibling of `aws-s3-multipart.mdx`, I wonder if there's a way to configure Docosorus to "do the right thing" here. In the mean time, let's add a trailing slash to reflect the runtime behavior.
```suggestion
[legacy version of this plugin](../aws-s3/). This is a suboptimal experience for
```

nqst: Using `../` is correct in this case. It's because we're moving between directories in the address bar. We start at `/docs/aws-s3-multipart/` and want to go to `/docs/aws-s3/`. To do this, we first go up to the parent directory, which is `docs`, and then into the `aws-s3` directory.

When we used `./`, the browser thought we were trying to find something inside the current directory, `/docs/aws-s3-multipart/`. But what we needed was to go up one level first, then into the correct directory. That's why `../` is the right choice.

Murderlon (approved):

Murderlon:
> Ideally, we should use server-side redirects instead. Is it possible?

No unfortunately not. GitHub Pages doesn't support it. We would have to switch to deploy on Netlify or Vercel I think.

arturi:
Thank you! All looks good to me except `pathname`. Are we better off with absolute urls? They will transfer better to a different engine if we switch from Docusaurus + self-explanatory.

nqst:
> Are we better off with absolute urls? They will transfer better to a different engine if we switch from Docusaurus + self-explanatory.

Good point! Just pushed an update.

PR #208, opened 2024-03-14
Make example page options persistent

### LINKED ISSUE #110: Uppy state and selected files should persist when options are changed on the /examples
image

### LINKED ISSUE #119: Checkboxes on the examples page need to be persistent
Otherwise when you click ✅ Golden Retriever, add a file and reload the page, nothing happens because the checkbox is unchecked again. You don't get to see the plugin in action.

Murderlon:
Closes #110 
Closes #119

Also fixes:
- Order of plugins are constantly changing when you check/uncheck them
- Screen Capture not working
- Golden Retriever never worked before since it would be off by default and it only shows what it does after a refresh. Persistent options fixes this too.
- Make Instagram / Facebook disabled with disabled styles for now (https://github.com/transloadit/uppy/issues/4905)

code is not pretty but it works and it's just a small page 🤷‍♂️

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,6 +55,7 @@
 		"@docusaurus/plugin-client-redirects": "^3.0.0",
 		"@docusaurus/preset-classic": "^3.0.0",
 		"@mdx-js/react": "^3.0.0",
+		"@uidotdev/usehooks": "^2.4.1",
 		"@uppy/audio": "latest",
 		"@uppy/box": "latest",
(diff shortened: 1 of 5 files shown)


lakesare (approved):
Great, thanks for taking a look at Golden Retriever too!
Shall we add `title="Our oauth app is currently getting validated by Facebook [Instagram for the instagram checkbox], but these options will work for your app"` to disabled checkboxes?

Issue #213, opened 2024-03-26
Use with your own server section of AWS S3 contains misleading routes


nerg4l:
The routes provided in the "[Use with your own server](https://uppy.io/docs/aws-s3-multipart/#use-with-your-own-server)" of AWS S3 uploader shows different endpoints than the endpoints used in the linked "[Setup Uppy](https://github.com/transloadit/uppy/blob/main/examples/aws-nodejs/public/index.html)" example.

https://github.com/transloadit/uppy.io/blob/6a5a3bc89cabd80e01b289e136783002ea9b38ab/docs/uploader/aws-s3-multipart.mdx?plain=1#L143-L151

From the [Setup Uppy example](https://github.com/transloadit/uppy/blob/6673b7787771194c52f0cb451bb1f3f69f8af71d/examples/aws-nodejs/public/index.html):

- `/sts`
- `/sign-s3`
- `/s3/multipart`
- `/s3/multipart/${uploadId}`
- `/s3/multipart/${uploadId}/${partNumber}`
- `/s3/multipart/${uploadId}/complete`

Murderlon:
They are the same, it just shows that if you have an existing backend, perhaps even with existing s3 endpoints , you can just group them under `/uppy` route of whatever you'd like to do.

nerg4l:
Are you sure that only the price is different?

- `/uppy/sts`
- `/uppy/s3`
- `/uppy/s3-multipart`
- `/uppy/s3-multipart/:id`
- `/uppy/s3-multipart/:id/:partNumber`
- `/uppy/s3-multipart/:id/complete`

The documentation uses `s3-multipart` but the linked code example uses `s3/multipart`. Also the sign route is completely different.

Murderlon:
Maybe the docs are misleading, but it just needs to match. You can do this if you want:

On the server:
```js
app.post('/my/supercool/endpoint', (req, res, next) => {}
```

And then on the client:

```js
const response = await fetch('/my/supercool/endpoint', {
  method: 'POST',
  // Send and receive JSON.
  headers: {
    accept: 'application/json',
  },
  body: serialize({
    filename: file.name,
    type: file.type,
    metadata,
  }),
  signal,
})
```

It's a lot of manual work. Ideally, we'd have `@uppy/server-functions` which you can use on the server, and force people to have the same endpoints as companion. Then I think you wouldn't need to manually write this.

nerg4l:
Agreed, a companion compatible uploader config would be even better.

[...]

PR #214, opened 2024-03-27
Modify endpoints in own server section of AWS S3

nerg4l:
The first PR to change the endpoints in the documentation.

https://github.com/transloadit/uppy/blob/960362b373666b18a6970f3778ee1440176975af/packages/%40uppy/companion/src/server/controllers/s3.js#L439-L447

Related to #213

### DIFF
--- a/docs/uploader/aws-s3-multipart.mdx
+++ b/docs/uploader/aws-s3-multipart.mdx
@@ -136,22 +136,9 @@ bucket in question. In-depth documentation about CORS rules is available on the
 The recommended approach is to integrate `@uppy/aws-s3` with your own server.
 You will need to do the following things:
 
-1. [Setup up a S3 bucket](#setting-up-your-s3-bucket).
-2. Create endpoints in your server. You can create them as edge functions (such
-   as AWS Lambdas), inside Next.js as an API route, or wherever your server
-   runs.
-   - `GET` > `/uppy/sts`: get the temporary security credentials (optional).
-   - `POST` > `/uppy/s3/params`: get parameters and pre-signed URL for non-multipart
-     upload.
-   - `POST` > `/uppy/s3/multipart`: create the multipart upload.
-   - `GET` > `/uppy/s3/multipart/:id`: get the uploaded parts.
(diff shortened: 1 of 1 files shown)


Murderlon (commented):
Thanks for making a PR.

Now that I think about it, does it make sense to duplicate endpoints of the example? I feel like it's just likely to get out of date again.

Some alternatives:
1. Instead of writing the endpoints and a link the example root, we can also create to steps:
   1. [Setup Uppy client](https://github.com/transloadit/uppy/blob/main/examples/aws-nodejs/public/index.html)
   2. [Setup your server](https://github.com/transloadit/uppy/blob/main/examples/aws-nodejs/index.js)
2. Or we keep the HTTP methods but remove the specific endpoints and add a sentence above that you need you need these endpoints but it's up to you to decide under which route they fall, as long as you keep it in sync with Uppy.

What do you think?

nerg4l:
I'm still getting familiar with the available Uppy components.

These endpoints will only change if `@uppy/companion` changes, am I wrong?

I think `@uppy/aws-s3-multipart` could be used for the "Setup Uppy client". That component has a predefined set of routes.

Also, `@uppy/companion` has great documentation for the routes. Example: https://github.com/transloadit/uppy/blob/960362b373666b18a6970f3778ee1440176975af/packages/%40uppy/companion/src/server/controllers/s3.js#L36-L50
Maybe it's enough to put these somewhere in the documentation.

Murderlon:
> These endpoints will only change if `@uppy/companion` changes, am I wrong?

No we are updating the section on using it with your own server. At the moment the endpoints don't matter at all. For instance in `createMultipartUpload` you have to use `fetch` to your own backend. You're in control what endpoint you fetch, not Uppy.

As discussed before we should change the property name from `companionUrl` to `endpoint` and document what to do _exactly_ so you don't have to write this all manually. But until this in there, we need to write docs for the current reality, brining me back to my two suggestions above.

nerg4l:
From a user's point of view, I think the first option would be better.

Murderlon:
Would you like to make the change?

nerg4l:
I will update this PR later today. I can only work on it outside of European working hours.

Murderlon (approved):

PR #227, opened 2024-05-05
Add dependabot config file

aduh95:
We are behind on some GHA as we can see from the warnings in the Actions tab. I'm also adding npm packages to the automation as it seems unlikely we're going to spend time maintaining it if the bots doesn't open the PRs for us.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,14 @@
+# Basic set up for three package managers
+
+version: 2
+updates:
+  # Maintain dependencies for GitHub Actions
+  - package-ecosystem: 'github-actions'
+    directory: '/'
+    schedule:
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,14 @@
...
+  # Maintain dependencies for GitHub Actions
+  - package-ecosystem: 'github-actions'
+    directory: '/'
+    schedule:
+      interval: 'weekly'
+
+  - package-ecosystem: 'npm'
+    directory: '/'
+    schedule:
+      interval: 'weekly'
```
Murderlon, line 14: I would also be okay with getting npm monthly to reduce the noise. 

```suggestion
      interval: 'monthly'
```

PR #229, opened 2024-05-06
feat: update home page vue example

LouisHaftmann:
I just updated the Vue example because it still used Vue's old options api.


Murderlon (approved):

Murderlon:
Thanks

PR #241, opened 2024-05-09
Add support for Svelte syntax highlighting

Murderlon:
Ref: https://github.com/transloadit/uppy/pull/5145


Murderlon:
To my surprise, this isn't supported.

aduh95:
FWIW I tried to patch prismjs to force it to count `svelte` as `markup` (like e.g. `html` or `svg`), but that didn't work. Giving up now

PR #293, opened 2024-12-18
add pickers to example

mifi:


Murderlon:
I think you need to `yarn install` them first

mifi:
oops good catch, will do. i just edited this in the github editor

Murderlon:
> i just edited this in the github editor

Have you tested locally this works as expected?

mifi:
yes but i didn't test the actual flow because of allowed origins, I just saw that the oauth window popped up

mifi:
let me test with my own key instead

mifi:
For Google photos it works. For drive I'm getting `The API developer key is invalid.` but it could be due to http://localhost

Murderlon:
> but it could be due to http://localhost

if they both have it in their allow list in the Google app, it should behave the same? Do you expect it will work suddenly when live? Or better to try something first before merging?

mifi:
i had incorrectly copy-pasted the api key locally, works now

Murderlon (approved):

PR #303, opened 2025-01-10
Update companion.md

mifi:


Murderlon (commented):
I don't think this belong here and should instead of documented on the google-drive-picker docs page.

mifi:
When I read the text:

> To do so each provider plugin must be configured with Transloadit’s Companion URLs:
> ...
> You may also hit rate limits, because the OAuth application is shared between everyone using Transloadit.

> To solve that, you can use your own OAuth keys with Transloadit’s hosted Companion servers by using Transloadit Template Credentials. [Create a Template Credential](https://transloadit.com/docs/#how-to-create-template-credentials) on the Transloadit site. Select “Companion OAuth” for the service, and enter the key and secret for the provider you want to use. Then you can pass the name of the new credentials to that provider:

I feel it it is written in a way that it applies to all Companion plugins, and so it can be confusing because what's written doesn't apply at all to Google Drive Picker and Google Photos Picker. that's why I wanted to mention it here.

Murderlon:
My assumption is that everyone who will need a google plugin is forced to go to the google plugin docs and thus they see it. This is more a generic mention of hosted Transloadit.

PR #305, opened 2025-01-10
Document google oauth2 situation

mifi:


Murderlon (changes requested):
I don't think this is the place for it. I would either make a blog post and link to it from the google plugins doc pages or create an MDX component and inline it in every plugin. This does not need to be a top sidebar item.

PR #318, opened 2025-02-04
improve picker docs

mifi:
to prevent known issues

### DIFF
--- a/docs/sources/companion-plugins/google-drive-picker.mdx
+++ b/docs/sources/companion-plugins/google-drive-picker.mdx
@@ -164,10 +164,9 @@ number_ under
 
 :::note
 
-Make sure you use the **project number** (**not** project ID or project
-name), and make sure you’re in the correct project. Note that if this is set to
-the wrong value it will be ignored by Google when your project is in testing
-mode, but once in production users will get 404 errors when trying to download.
+Make sure you use the **project number** (**not** project ID or project name),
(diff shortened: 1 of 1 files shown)


docs/sources/companion-plugins/google-drive-picker.mdx
```diff
@@ -146,8 +158,11 @@ The API key from the [Initial setup](#initial-setup) (`string`).
...
-`Project number` under `IAM & Admin` > `Settings`. Note that if this is set to
-an invalid value it will be ignored by Google when your project is in testing
-mode, but once in production you will get 404 errors when trying to download
+[Google Developer Console](https://console.developers.google.com): _Project
+number_ under
+[`IAM & Admin` > `Settings`](https://console.cloud.google.com/iam-admin/settings).
+Warning: make sure you use the **project number** (**not** project ID or project
+name), and make sure you’re in the correct project. Note that if this is set to
+the wrong value it will be ignored by Google when your project is in testing
+mode, but once in production users will get 404 errors when trying to download
```
Murderlon, line 167: We have admonitions for this. I would recommend `note` or `caution`

```suggestion

:::note

Make sure you use the **project number** (**not** project ID or project
name), and make sure you’re in the correct project. Note that if this is set to
the wrong value it will be ignored by Google when your project is in testing
mode, but once in production users will get 404 errors when trying to download.

:::
```

PR #320, opened 2025-02-10
Update google-drive-picker.mdx

mifi:
see https://github.com/transloadit/uppy/issues/5625

### DIFF
--- a/docs/sources/companion-plugins/google-drive-picker.mdx
+++ b/docs/sources/companion-plugins/google-drive-picker.mdx
@@ -73,7 +73,9 @@ Using Google Drive Picker requires setup in both Uppy and Companion.
 
 ### Prerequisites
 
-Make sure your web app is hosted on a server that either does **not set `Cross-Origin-Opener-Policy` at all** or sets it to `same-origin-allow-popups`. [See this issue for more information](https://github.com/transloadit/uppy/issues/5625).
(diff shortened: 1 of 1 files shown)


docs/sources/companion-plugins/google-drive-picker.mdx
```diff
@@ -71,6 +71,12 @@ yarn add @uppy/google-drive-picker
 
 Using Google Drive Picker requires setup in both Uppy and Companion.
 
+### Prerequisites
+
+Make sure your web app is hosted on a server that either does **not set
+`Cross-Origin-Opener-Policy` at all** or sets it to `same-origin-allow-popups`.
+[See this issue for more information](https://github.com/transloadit/uppy/issues/5625).
```
Murderlon, line 78: Is there more relevant information or is this basically it? Ideally we don't have to link it if it doesn't add much

Fingel: I'm the author of the linked bug. It probably doesn't add much more information, other than  Django by default sets the header to `same-origin` which breaks the plugin, completely silently and inexplicably. It was a real head-scratcher to figure out.

Murderlon:
```suggestion
```

Murderlon (approved):

PR #332, opened 2025-03-20
Make comparison table

mifi:


Murderlon (changes requested):
I would expect comparison to be a docs menu item, not on the front page. That's where most products I know put it. People mostly skipping the homepage will never see it this way. 

If we make it a docs page we have more room too, then I would consider making two tables, one for sources as it's currently very repetitive just for that, and one for features. For the features one we should include:
-  some things we don't do well on as well, such as headless components, server-side integration, server-side restrictions, access control
- framework support (if we also include meta-frameworks, then separate table as well)
- different uploaders
- client-side restrictions
- forms integration
- i18n
- compressor
- advanced provider views UX. See [these examples](https://uppy.io/blog/uppy-4.0/#ux-improvements-for-viewing-remote-files).
- 


Lastly, "paid hosted option" is confusing. Uppy doesn't have that, Transloadit is something else. I would remove it. Or rename it and explicitly link to Transloadit.

mifi:
please have another look :) i've addressed most of the feedback. some of the new fields I haven't researched yet, so they are all question mark, but I can look into that if you think it looks good.

i'm a bit unsure about the "uploaders" part, because it's mostly only Uppy/Companion that can upload to those destinations directly. The other uploader services require the client to upload to **their** proprietary hosted solution (kind of like using uppy with Transloadit plugin). And in that case, usually the sky is the limit. Like how with Transloadit(assemblies) you can upload to pretty much any destination thinkable, and it doesn't make sense to list all of them.

PR #339, opened 2025-04-04
dropbox docs: explain team folders

mifi:
add notes about team folders

see also https://github.com/transloadit/uppy/pull/5708

### DIFF
--- a/docs/sources/companion-plugins/dropbox.mdx
+++ b/docs/sources/companion-plugins/dropbox.mdx
@@ -127,9 +127,10 @@ Things to note:
 - Choose the “Dropbox API”, not the business variant.
 - Typically you’ll want “Full Dropbox” access, unless you are absolutely certain
   that you need the other one.
-- If you want to support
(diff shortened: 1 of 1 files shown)


docs/sources/companion-plugins/dropbox.mdx
```diff
@@ -127,6 +127,9 @@ Things to note:
 - Choose the “Dropbox API”, not the business variant.
 - Typically you’ll want “Full Dropbox” access, unless you are absolutely certain
   that you need the other one.
+- If you want to support
+  [Team folders / Teams spaces](https://developers.dropbox.com/dbx-team-files-guide),
+  counterintuitively you must **not** enable any Team scopes under Permissions.
```
Murderlon, line 132: Shouldn't we say then what you should do? The phrasing makes it sound like something is missing?

mifi: good point. i've updated and added instructions aobut which scopes are needed now

Murderlon (approved):

PR #343, opened 2025-05-15
Added stackblitz embeds

qxprakash:
@Murderlon, the embed URLs in this PR are from my modified fork because currently, almost all examples use dependencies from the workspace. , I'm thinking of a potential workaround to this

image


Murderlon:
We should change all the dependencies to not use the workspace indeed. 

Let's wait till headless components is merged before merging this.

Murderlon (approved):
I applied some changes as well

PR #354, opened 2025-06-11
Fix documentation for google drive

mifi:
Actually this might be because the app is not verified. Maybe Once the app is verified, Google will show the correct name

### DIFF
--- a/docs/sources/companion-plugins/google-drive.mdx
+++ b/docs/sources/companion-plugins/google-drive.mdx
@@ -103,10 +103,8 @@ uppy.use(GoogleDrive, {
 });
 ```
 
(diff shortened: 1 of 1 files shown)


Murderlon (approved):
All good once CI is green

mifi:
how to auto fix issues like this? `yarn remark:mdx` doesn't seem to do anything

Murderlon:
I don't understand what's so difficult, CI clearly points to the line where you are using the wrong quotes. I just pushed a commit.

[Using the correct quotes is good typography hygiene](https://practicaltypography.com/straight-and-curly-quotes.html).

PR #375, opened 2025-08-25
Update paths to style.css

fredrikcarlbom:
Update paths to style.css and style.min.css
The following are not updated since they don't work drag-drop, drag-target, file-input, informer, progress-bar and status-bar.


Murderlon:
Thanks a lot for the PR, but I already #374 which also includes some other fixes :)

PR #380, opened 2025-09-11
update docs post 5.0 release

qxprakash:
- Remove docs for deprecated plugins and their references: , `@uppy/informer` , `@uppy/file-input` , `@uppy/progress-bar`
- We should also cleanup -- https://uppy.io/docs/guides/custom-stores/#reduxstore , as `@uppy/redux` and `@uppy/redux-devtools`were also removed in https://github.com/transloadit/uppy/pull/5834

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -4958,14 +4958,14 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@uppy/dashboard@npm:5.0.3, @uppy/dashboard@npm:^5.0.0, @uppy/dashboard@npm:^5.0.2":
-  version: 5.0.3
-  resolution: "@uppy/dashboard@npm:5.0.3"
+"@uppy/dashboard@npm:5.0.4, @uppy/dashboard@npm:^5.0.0, @uppy/dashboard@npm:^5.0.2":
+  version: 5.0.4
+  resolution: "@uppy/dashboard@npm:5.0.4"
   dependencies:
     "@transloadit/prettier-bytes": "npm:^0.3.4"
-    "@uppy/provider-views": "npm:^5.1.1"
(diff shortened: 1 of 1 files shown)


qxprakash:
should we also mention that the plugin is deprected in framework docs ? 

for example :

https://github.com/transloadit/uppy.io/blob/fbc3f8591e6319cfb291fef08d72110aa91d84ac/docs/framework-integrations/react.mdx#L56-L61

something like -`<ProgressBar />` renders [`@uppy/progress-bar`](/docs/progress-bar) — Deprecated in Uppy 5.0`

docs/sources/file-input.mdx
```diff
@@ -10,6 +10,15 @@ import UppyCdnExample from '/src/components/UppyCdnExample';
 
 # File input
 
+::::caution Deprecated
```
Murderlon: did you double check putting "Deprecated" on the same line renders as expected?

docs/sources/file-input.mdx
```diff
@@ -10,6 +10,15 @@ import UppyCdnExample from '/src/components/UppyCdnExample';
...
+
+As of Uppy 5.0, `@uppy/file-input` is deprecated. Prefer the headless hooks
+(e.g., `useDropzone`) from `@uppy/react`, `@uppy/vue`, or `@uppy/svelte`, or use
```
Murderlon:
```suggestion
(e.g., `useFileInput`) from `@uppy/react`, `@uppy/vue`, or `@uppy/svelte`, or use
```

qxprakash:
@Murderlon should we merge this ?

Murderlon (commented):
Framework docs are missing and the language is not clear, "deprecated" means it's still in the bundle, while in fact we completely removed it.

PR #422, opened 2026-02-18
Update note on UppyContextProvider usage

mifi:
Clarified usage of UppyContextProvider in React components.

because when using AI agent, it tried to place multiple context providers around the code.

### DIFF
--- a/docs/framework-integrations/react.mdx
+++ b/docs/framework-integrations/react.mdx
@@ -71,7 +71,8 @@ Play around with the React headless components in StackBlitz
 
 :::note
 
-All components must be placed within a `UppyContextProvider`.
+All components must be placed within a `UppyContextProvider`. You typically only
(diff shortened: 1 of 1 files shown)


Murderlon (commented):
This doesn't add anything imho. It's common knowledge that context providers are only meant to be used _once_. If AI made this mistake once, doesn't mean it will make it consistently 🤔

mifi:
I would argue that an LLM **is** common knowledge because it has been trained on common knowledge, and if it makes the mistake then it's likely that common knowledge doesn't say that there must only be a single context provider. and there are cases where context providers are used many times, like for example when using UI libraries and themes need to be overridden for a subtree.

Murderlon:
I'm fairly certain it was a fluke, working on evaluations myself and running LLMs on the same tasks dozens of times, the output is very different consistently, I wouldn't change docs based on one fluke personally.

